PR #11, opened 2022-03-10
fix: yarn v3

vuvoth:
Fix error when use yarn install.


package.json
```diff
@@ -37,5 +37,6 @@
 	},
 	"dependencies": {
 		"ant-design-vue": "next"
-	}
+	},
+	"packageManager": "yarn@3.2.0"
```
robert-zaremba, line 41: :+1:

robert-zaremba (approved):
Thanks @vuvth

PR #20, opened 2022-03-15
WebPack 5 polyfills required added

rmeena840:
WebPack 5 polyfills required added
Dependency like crypto, buffer etc has been added


robert-zaremba (approved):

robert-zaremba:
There are lot of outdated dependencies, let's update them. Could you do it?

PR #21, opened 2022-03-15
Use algo-builder/web in connect wallet functions

rmeena840:
Use algo-builder/web in connect wallet functions

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -5,7 +5,7 @@
 <script lang="ts">
 import { defineComponent } from "vue";
 import WalletStore from "./store/WalletStore";
-import { Wallet } from "./types";
+import { WalletType } from "./types";
 import { MAX_WALLET_COUNT_CHECK } from "./config/app.config";
 declare var AlgoSigner: any; // eslint-disable-line
 
@@ -20,7 +20,7 @@ export default defineComponent({
 		const walletStore = WalletStore();
 
 		return {
-			setHasAlgoSigner: walletStore.setHasAlgoSigner,
+			setWalletType: walletStore.setWalletType,
 		};
 	},
 	mounted() {
@@ -29,7 +29,7 @@ export default defineComponent({
 	methods: {
 		async checkAlgoSigner() {
 			if (typeof AlgoSigner !== "undefined") {
-				this.setHasAlgoSigner(Wallet.ALGOSIGNER);
+				this.setWalletType(WalletType.ALGOSIGNER);
 				await AlgoSigner.connect();
 			} else {
 				// maximum number of times to check wallet
--- a/src/App.vue
+++ b/src/App.vue
@@ -11,17 +11,16 @@ declare var AlgoSigner: any; // eslint-disable-line
 
 export default defineComponent({
 	name: "App",
-	components: {},
 	data() {
 		return {
 			walletCheckCount: 1,
 		};
 	},
 	setup() {
-		const counter = WalletStore();
+		const walletStore = WalletStore();
 
 		return {
-			setHasAlgoSigner: counter.setHasAlgoSigner,
+			setHasAlgoSigner: walletStore.setHasAlgoSigner,
 		};
 	},
 	mounted() {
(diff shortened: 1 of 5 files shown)


src/App.vue
```diff
@@ -11,17 +11,16 @@ declare var AlgoSigner: any; // eslint-disable-line
...
 			walletCheckCount: 1,
 		};
 	},
 	setup() {
-		const counter = WalletStore();
+		const walletStore = WalletStore();
 
 		return {
-			setHasAlgoSigner: counter.setHasAlgoSigner,
+			setHasAlgoSigner: walletStore.setHasAlgoSigner,
```
robert-zaremba: Why we are using "set algo signer"? What if a user will like to use WalletConnect.

I think we should have a function `walletStore.connect(type, ...)`, where type is something like (I'm sure we already have some enum for that):
```
enum WalletType {
   WalletConnect
   AlgoSigner
   MyAlgo
}
```

PS: if this requires an update in the web package, then please add appropriate tasks (one for the web package and one task for the sigma dao). This is a high priority - because we want to release the web package soon. so we need to make sure we have all required API to connect to all wallets.

src/components/WalletConnect.vue
```diff
@@ -43,9 +43,17 @@ export default defineComponent({
 			walletAddress: "Not Found",
 		};
 	},
+	setup() {
+		const walletStore = WalletStore();
+		return {
+			initializeWebMode: walletStore.setWebMode,
+		};
+	},
```
robert-zaremba, line 51: why we have two `walletStore`s? this one looks more reasonable than the one in App.vue (it is using the `setWebMode` rather than `setHasAlgoSigner`). Could you explain why we have two stores, and which one we should use?

src/store/WalletStore.ts
```diff
@@ -1,15 +1,20 @@
...
+			webMode: <WebMode>{},
 		};
 	},
 	actions: {
 		setHasAlgoSigner(value: Wallet) {
 			this.walletKind = value;
 		},
+		setWebMode(value: WebMode) {
+			console.log("WebMode Initialized", value);
+		},
```
robert-zaremba:
```suggestion
		setWebMode(m: WebMode) {
			console.log("WebMode Initialized", m);
		},
```

robert-zaremba: `value` variable name is too generic in this context, we can find better names.

robert-zaremba (commented):
looks good, I left few questions.

src/App.vue
robert-zaremba: hmm, maybe we should remove `walletStore.setHasAlgoSigner` function?

src/components/WalletConnect.vue
rmeena840: walletStore is instance of store which is used for connecting to store from WalletConnect.vue file

src/App.vue
rmeena840: This enum already exists. Yes, you are right I should give the right function name and make it setWalletType

src/store/WalletStore.ts
rmeena840: Ok. I will make it webMode

Megha-Dev-19 (commented):
Lgtm

src/App.vue
robert-zaremba: Exactly, that's why I wrote that _I'm sure we already have ...._

src/components/WalletConnect.vue
robert-zaremba: yes, but why we have two instance? On here and one in App.vue?

src/store/WalletStore.ts
robert-zaremba: I think we don't need to repeat the type, `m` or `mode` is enough, but as  you prefer.

Megha-Dev-19 (approved):

src/types/enum.types.ts
```diff
@@ -1,4 +1,4 @@
-export enum Wallet {
+export enum WalletType {
 	NONE,
 	ALGOSIGNER,
 	MY_ALGO,
```
thdailong, line 4: Can you explain for me a bit about the enum Wallet?

thdailong (approved):
Okay

PR #22, opened 2022-03-16
Added menu, proposal form, vote form, Dao ID input

Megha-Dev-19:
image

image

### DIFF
--- a/src/components/AppHeader.vue
+++ b/src/components/AppHeader.vue
@@ -1,37 +1,52 @@
 <template>
+	<!-- TODO: Update to ANTD menu and header -->
 	<div class="header" id="header">
-		<div class="grid">
-			<div class="flexbox">
-				<h1>SigmaDao</h1>
-				<div class="menu_id">
-					<label for="dao_id" style="padding-right: 10px">DAO ID</label>
-					<input id="dao_id" v-model="id" type="number" />
+		<a-row class="veritcal_center">
+			<a-col :span="20">
+				<a-row class="veritcal_center">
+					<a-col :span="4">
+						<h1>SigmaDao</h1>
+					</a-col>
+					<a-col :span="8">
+						<div class="menu_id">
+							<label for="dao_id" style="padding-right: 10px">DAO ID</label>
+							<input id="dao_id" v-model="id" type="number" />
+						</div>
+					</a-col>
+					<a-col>
+						<div>
+							<router-link :to="{ path: '/vote' }">
+								<a-button
+									class="menu_option"
+									:type="
+										currentPageKey === NavigationKey.VOTE ? 'link' : 'text'
+									"
+									@click="() => handleMenuClick(NavigationKey.VOTE)"
+									>Vote</a-button
+								>
+							</router-link>
+							<router-link :to="{ path: '/addProposal' }">
+								<a-button
+									class="menu_option"
+									:type="
+										currentPageKey === NavigationKey.ADD_PROPOSAL
+											? 'link'
+											: 'text'
+									"
+									@click="() => handleMenuClick(NavigationKey.ADD_PROPOSAL)"
+									>Add Proposal</a-button
+								>
+							</router-link>
+						</div>
+					</a-col>
+				</a-row>
+			</a-col>
+			<a-col :span="4">
+				<div>
+					<WalletConnect />
 				</div>
-				<div class="menu">
-					<router-link :to="{ path: '/vote' }">
-						<a-button
-							class="menu_option"
-							:type="currentPageKey === NavigationKey.VOTE ? 'link' : 'text'"
-							@click="() => handleMenuClick(NavigationKey.VOTE)"
-							>Vote</a-button
-						>
-					</router-link>
-					<router-link :to="{ path: '/addProposal' }">
-						<a-button
-							class="menu_option"
-							:type="
-								currentPageKey === NavigationKey.ADD_PROPOSAL ? 'link' : 'text'
-							"
-							@click="() => handleMenuClick(NavigationKey.ADD_PROPOSAL)"
-							>Add Proposal</a-button
-						>
-					</router-link>
-				</div>
-			</div>
-			<div class="right_corner">
-				<WalletConnect />
-			</div>
-		</div>
+			</a-col>
+		</a-row>
 	</div>
 </template>
 
@@ -82,31 +97,18 @@ export default defineComponent({
 	font-size: 15px !important;
 }
 
-.grid {
-	display: grid;
-	grid-template-columns: auto max-content;
+.veritcal_center {
 	align-items: center;
-}
-
-.right_corner {
-	justify-self: flex-end;
 }
 
 .flexbox {
 	display: flex;
 }
 
-.menu {
-	position: absolute;
-	left: 50%;
-	transform: translateX(-50%);
-	align-self: center;
-}
 .menu_option {
 	font-size: 20px;
 }
 .menu_id {
 	align-self: center;
-	margin-inline: 20px;
 }
 </style>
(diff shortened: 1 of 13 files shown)


src/components/AppHeader.vue
```diff
@@ -1,5 +1,112 @@
...
+.flexbox {
+	display: flex;
+}
+
+.menu {
+	position: absolute;
+	left: 50%;
+	transform: translateX(-50%);
+	align-self: center;
+}
```
robert-zaremba: please add TODO: we will need to remove it when we will use `antd.Menu` component.

src/components/AppHeader.vue
```diff
@@ -1,5 +1,112 @@
...
+	background-color: white;
+	z-index: 9998;
+	font-size: 15px !important;
+}
+
+.grid {
+	display: grid;
+	grid-template-columns: auto max-content;
+	align-items: center;
+}
```
robert-zaremba, line 96: why not using Grid component from antd?

src/components/AppHeader.vue
```diff
@@ -1,5 +1,112 @@
...
+	box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.1);
+	position: fixed;
+	top: 0;
+	left: 50%;
+	transform: translateX(-50%);
+	width: 1400px;
+	background-color: white;
+	z-index: 9998;
+	font-size: 15px !important;
+}
```
robert-zaremba, line 92: we should use `Header` and [`Layout`](https://www.antdv.com/components/layout/) component from atnd. We can do it in the UI Layout task (in that case please add a TODO comment here) or now if you prefer.

Goal: avoid custom too many styles.

src/components/AppHeader.vue
```diff
@@ -1,5 +1,112 @@
...
+});
+</script>
+<style scoped>
+.header {
+	padding-block: 10px;
+	padding-inline: 20px;
+	height: auto;
+	box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.1);
+	position: fixed;
+	top: 0;
```
robert-zaremba: we should not use fixed positining in header. This is why they wallet switch doesn't work.

src/components/WalletConnect.vue
```diff
@@ -61,11 +58,23 @@ export default defineComponent({
...
 			}
 		},
-		handleMenuClick(e: Event) {
-			console.error("changing wallet kind", e);
+		handleMenuClick(e: any) {
+			console.error("changing wallet kind", e.key);
+			if (e.key === "1") {
+				this.selectedWallet = "1";
+				this.connectAlgoSigner();
+			}
```
robert-zaremba, line 83: Add : 
```
else { console.warn("Wallet %s not supported", e.key); 
```

src/components/WalletConnect.vue
```diff
@@ -61,11 +58,23 @@ export default defineComponent({
...
 			if (userAccount && userAccount.length) {
 				this.walletAddress = userAccount[0].address;
+				this.text = "AlgoSigner";
 			}
 		},
-		handleMenuClick(e: Event) {
-			console.error("changing wallet kind", e);
+		handleMenuClick(e: any) {
+			console.error("changing wallet kind", e.key);
+			if (e.key === "1") {
```
robert-zaremba: we must have an enum , something like `WalletKind.AlgoSigner`. I think @rmeena840  was adding it somewhere.

src/pages/Proposal.vue
```diff
@@ -0,0 +1,174 @@
...
+			<div class="flexBox" v-if="formState.vote_type === ProposalType.ASA">
+				<label for="asa_id">ASA ID</label>
+				<input id="asa_id" v-model="formState.asaId" />
+			</div>
+			<div class="flexBox" v-if="formState.vote_type === ProposalType.MESSAGE">
+				<label for="message">Message</label>
+				<input id="message" v-model="formState.message" />
+			</div>
+			<button type="submit" class="submit_btn" @click="onSubmit">Submit</button>
+		</form>
```
robert-zaremba: why not using https://www.antdv.com/components/form/ ?

src/pages/Proposal.vue
```diff
@@ -0,0 +1,174 @@
...
+						return (success = formState.asaId ? true : false);
+					case ProposalType.MESSAGE:
+						return (success = formState.message ? true : false);
+					default:
+						return (success =
+							formState.from || formState.recipient || formState.amount
+								? true
+								: false);
+				}
+			}
```
robert-zaremba: add a proper error when something is wrong. In fact https://www.antdv.com/components/form/ handles nicely error reporting.
We could also use: https://www.antdv.com/components/alert/

robert-zaremba (commented):
+ Don't display address if the wallet is not connected (instead of showing "Address not found")
+ let's use more of the antd components - they have nice styling and solve other problems (like verification, feedback...)

src/components/AppHeader.vue
Megha-Dev-19: okay

src/components/AppHeader.vue
Megha-Dev-19: I didn't know that we have to minimise own custom styling. In future will keep that in mind. Thanks for the update.

src/components/AppHeader.vue
Megha-Dev-19: okay will add a TODO for now, will update the menu later as per the task you have created.

src/components/AppHeader.vue
Megha-Dev-19: but to make it sticky, I have to use either `fixed/absolute/sticky`, but since we will use antd menu, I guess that would solve the problem.

src/pages/Proposal.vue
Megha-Dev-19: I tried using it, but its validation and error handling is not working as expected. I also had a call with @rmeena840 regarding it but we couldn't find any fix. So I switched to basic html, we can discuss about it later if you want.

src/pages/Proposal.vue
Megha-Dev-19: As I mentioned earlier, antd form isn't working well, but I can use alert to show errors. Will do that

src/components/WalletConnect.vue
Megha-Dev-19: yes, that would be much better. Thanks for informing

src/pages/Proposal.vue
robert-zaremba: OK, let's leave it for now. In the past I successful used antd form and it worked well

robert-zaremba: I've created a task for it: https://www.pivotaltracker.com/story/show/181590936

src/components/AppHeader.vue
robert-zaremba: That's why we are using Antd - to reuse pre made components rather than spending time for creating our own ones.

src/components/AppHeader.vue
robert-zaremba: but we don't need to make it sticky

src/pages/Proposal.vue
robert-zaremba: BTW: currently we don't have error validation either. Anyway - let's migrate to antd from in a new task.

robert-zaremba (approved):

src/pages/Proposal.vue
Megha-Dev-19: I guess there was some version problem along with new keys being added, I have updated the form to antd form.

src/components/AppHeader.vue
Megha-Dev-19: okay, then we don't need it. Will update

PR #23, opened 2022-03-16
Upgrade dependency for sigmadao

rmeena840:
Upgrade dependency for sigmadao


package.json
```diff
@@ -15,42 +15,45 @@
...
 	},
 	"dependencies": {
 		"@algo-builder/algob": "^3.2.0",
 		"@algo-builder/web": "^3.2.0",
-		"@ant-design/icons-vue": "^6.0.1",
+		"@ant-design/icons-vue": "^6.1.0",
 		"algosdk": "^1.14.0",
-		"ant-design-vue": "3.0.0-beta.13",
-		"core-js": "^3.6.5",
+		"ant-design-vue": "next",
```
robert-zaremba: Why do you use "next"? you must not use "next" here. It will be resolved to v2.x.... which is based on the old vue.

Instead, please use: `3.1.0-rc.01`

rmeena840: yarn install gives an error that this version is not found. 3.1.0-rc.0 is available and it working. I can use that.

robert-zaremba (approved):

PR #25, opened 2022-03-23
Update Wallet component and disconnect feature

rmeena840:
Update Wallet component and disconnect feature

### DIFF
--- a/src/components/AppHeader.vue
+++ b/src/components/AppHeader.vue
@@ -43,9 +43,13 @@
 			</a-col>
 			<a-col :span="4">
 				<div>
-					<WalletConnect />
+					<WalletConnect @updateWalletAddress="updateWalletAddress($event)" />
 				</div>
 			</a-col>
+		</a-row>
+		<a-row v-if="walletAddress" justify="end">
+			Address: <br />
+			{{ walletAddress }}
 		</a-row>
 	</div>
 </template>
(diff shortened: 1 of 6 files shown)


src/pages/Proposal.vue
```diff
@@ -136,12 +136,13 @@ export default defineComponent({
 	setup() {
 		const formState = ProposalStore();
 
-		const onFinish = (values: any) => {
+		const onFinish = (values: Event) => {
```
robert-zaremba, line 139: :+1: , we should really avoid using `any`

src/components/AppHeader.vue
```diff
@@ -43,10 +43,14 @@
...
 			<a-col :span="4">
 				<div>
-					<WalletConnect />
+					<WalletConnect @updateWalletAddress="updateWalletAddress($event)" />
 				</div>
 			</a-col>
 		</a-row>
+		<a-row v-if="walletAddress" justify="end">
+			Address: <br />
+			{{ walletAddress }}
```
robert-zaremba, line 52: why breaking it in 2 lines?

robert-zaremba:
@Megha-Dev-19 Could you check if it looks good, and do a final approval?

src/components/AppHeader.vue
rmeena840: It's a line break. It will look something like the below. I felt like this would look good.
Address:
<<the address will come here>>

src/pages/Proposal.vue
rmeena840: Agreed

Megha-Dev-19 (approved):

PR #26, opened 2022-03-28
Header UI layout

Megha-Dev-19:
- Improved header layout using antd menu and header components.
- Made the layout responsive.
- added address to wallet store as it can be used across pages
- added logo (using image converted from svg since svg was taking more space and showing smaller logo)

image

image


src/components/AppHeader.vue
```diff
@@ -1,56 +1,72 @@
...
+								>
+								<router-link :to="{ path: '/vote' }">
+									<a-button
+										class="menu_option"
+										:type="isLinkActive(NavigationKey.VOTE)"
+										@click="() => handleMenuClick(NavigationKey.VOTE)"
+										>Vote</a-button
+									>
+								</router-link>
+								<router-link :to="{ path: '/addProposal' }">
```
rmeena840: We can use enum here. If in future if we decide to change the name of the endpoint then there will be very less code change.

e.g
enum Pagination {
ADD_PROPOSAL: '/addProposal',
VOTE: '/vote'
}

src/components/AppHeader.vue
```diff
@@ -66,18 +82,19 @@ export default defineComponent({
...
+		handleMenuClick(value: number) {
 			this.currentPageKey = value;
 			console.log(value);
 		},
-		updateWalletAddress(walletAddress: string) {
-			this.walletAddress = walletAddress;
+		isLinkActive(currentPage: number) {
+			if (this.currentPageKey === currentPage) {
+				return "link";
+			} else return "text";
```
rmeena840:
```suggestion
			} return "text";
```

rmeena840 (commented):
Have left a few comments.

src/components/AppHeader.vue
Megha-Dev-19: nice suggestion. will update

robert-zaremba (commented):
1. Let's remove the border from the header.
2. There is already logo.svg - let's use it instead of creating the png file. Especially that it looks different.
3. can we set max width to the DAO info table to something like 1000px?
4. Does it display correctly on small screen?

src/App.vue
```diff
@@ -14,3 +16,31 @@ export default defineComponent({
...
+
+@media only screen and (min-width: 112.5em) {
+	html {
+		font-size: 75%;
+	}
+}
+.app_root {
+	max-width: 120rem;
+	margin: auto;
+}
```
robert-zaremba: shall we have a proper CSS file for that? I'm sure there is a proper way to customize antd styles and have a global style definitions to include.

src/components/AppHeader.vue
```diff
@@ -1,63 +1,79 @@
...
-											? 'link'
-											: 'text'
-									"
-									@click="() => handleMenuClick(NavigationKey.ADD_PROPOSAL)"
-									>Add Proposal</a-button
+									:type="isLinkActive(NavigationKey.ALL_DAO)"
+									@click="() => handleMenuClick(NavigationKey.ALL_DAO)"
+									>All DAOs</a-button
+								>
+								<router-link :to="{ path: EndPoint.VOTE }">
```
robert-zaremba: please rename `EndPoint.VOTE` to `EndPoint.PROPOSALS`

src/components/AppHeader.vue
```diff
@@ -1,63 +1,79 @@
...
+									<a-button
+										class="menu_option"
+										:type="isLinkActive(NavigationKey.VOTE)"
+										@click="() => handleMenuClick(NavigationKey.VOTE)"
```
robert-zaremba: same here

src/components/AppHeader.vue
```diff
@@ -1,63 +1,79 @@
...
+	<div class="components-page-header-demo-responsive">
+		<a-page-header>
+			<template #title>
+				<img src="../assets/logo.png" class="logo" />
+			</template>
+			<template #extra>
+				<WalletConnect />
+			</template>
+			<div class="content">
+				<div class="main">
```
robert-zaremba: why do we have 3 nested divs (components-page-header-demo-responsive, content, main) + <a-page-header>?
Can we reduce it to 2?

So, basically we have 4 elements without anything in the middle, slike:
```
<div>  // components-page-header-demo-responsive
  <div>  // a-page-header
    <div>  // content
      <div> //main

         real content

      <div>
    <div>
  <div>
<div>
```

src/components/AppHeader.vue
```diff
@@ -89,25 +108,29 @@ export default defineComponent({
...
+	max-width: 1000px;
+	margin: auto;
+}
+.menu {
+	text-align: center;
+	margin-bottom: 1rem;
 }
 .menu_option {
-	font-size: 20px;
+	font-size: 1.8rem !important;
```
robert-zaremba: let's use LESS or SCSS variable. We should avoid creating ad hock custom font size. Instead create varibles, like: 
```
f-size1  = 1.2rem, 
f-size2  = ....
```

src/components/WalletConnect.vue
```diff
@@ -11,22 +11,40 @@
...
+							<a-button type="primary" @click="handleLogOut">
+								<template #icon>
+									<LogoutOutlined />
+								</template>
+							</a-button>
+						</a-col>
+					</a-row>
+				</a-card>
+			</a-col>
+		</a-row>
```
robert-zaremba, line 44: do we need both row and col?

Megha-Dev-19: No, since only one column is there, we can just have a row. will update

Megha-Dev-19:
I have added css files rather than adding styling to vue files for more cleaner and maintainable code

src/components/AppHeader.vue
```diff
@@ -1,63 +1,70 @@
...
+							<a-button
+								class="menu_option"
+								:type="isLinkActive(NavigationKey.CREATE_DAO)"
+								@click="() => handleMenuClick(NavigationKey.CREATE_DAO)"
+								>Create DAO</a-button
+							>
+
+							<a-button
+								class="menu_option"
+								:type="isLinkActive(NavigationKey.ALL_DAO)"
```
robert-zaremba: please rename ALL_DAO to DAOs

robert-zaremba (approved):
pre-approving. 
@thdailong can you do a review as well?

thdailong (approved):
Approved.

PR #27, opened 2022-03-29
Use data from the indexer to get header information

rmeena840:
Use data from the indexer to get header information

### DIFF
--- a/src/components/WalletConnect.vue
+++ b/src/components/WalletConnect.vue
@@ -36,6 +36,7 @@ import { defineComponent } from "vue";
 import { WalletType } from "../types/enum.types";
 import { WebMode } from "@algo-builder/web";
 import WalletStore from "../store/WalletStore";
+import { searchForAssets } from "../indexer";
 declare var AlgoSigner: any; // eslint-disable-line
 
 export default defineComponent({
@@ -49,6 +50,25 @@ export default defineComponent({
 			selectedWallet: WalletType.NONE,
 			WalletType,
 		};
+	},
+	mounted() {
+		searchForAssets("gov-token")
+			.then((response) => {
+				if (response && response.assets && response.assets.length) {
+					let assetLength: number = response.assets.length;
+					if (response.assets[assetLength - 1].params) {
+						console.log(
+							"Gov Token Info:",
+							response.assets[assetLength - 1].params
+						);
+					}
+				} else {
+					console.warn("gov-token not found");
+				}
+			})
+			.catch((error) => {
+				console.log(error);
+			});
 	},
 	setup() {
 		const walletStore = WalletStore();
--- a/src/components/WalletConnect.vue
+++ b/src/components/WalletConnect.vue
@@ -37,6 +37,7 @@ import { WalletType } from "../types/enum.types";
 import { WebMode } from "@algo-builder/web";
 import WalletStore from "../store/WalletStore";
 import { searchForAssets } from "../indexer";
+import { GOV_TOKEN_ASSET } from "../constants/constant";
 declare var AlgoSigner: any; // eslint-disable-line
 
 export default defineComponent({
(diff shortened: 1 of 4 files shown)


rmeena840:
@Megha-Dev-19 @thdailong @robert-zaremba  Please review this PR.

rmeena840:
@Megha-Dev-19 @thdailong @robert-zaremba  Please review this PR.

src/components/WalletConnect.vue
```diff
@@ -50,6 +51,25 @@ export default defineComponent({
 			WalletType,
 		};
 	},
+	mounted() {
+		searchForAssets("gov-token")
```
ratik21: From where do we get the name `"gov-token"`? Is the ASA Name (always constant)? If so, maybe we should keep it somewhere in `constants.ts` ?

src/components/WalletConnect.vue
```diff
@@ -36,6 +36,7 @@ import { defineComponent } from "vue";
 import { WalletType } from "../types/enum.types";
 import { WebMode } from "@algo-builder/web";
 import WalletStore from "../store/WalletStore";
+import { searchForAssets } from "../indexer";
 declare var AlgoSigner: any; // eslint-disable-line
```
ratik21, line 41: avoid using `any`. Can we import types from `Algosigner` repo? I think it's been moved to typescript

src/components/WalletConnect.vue
```diff
@@ -50,6 +51,25 @@ export default defineComponent({
...
+				if (response && response.assets && response.assets.length) {
+					let assetLength: number = response.assets.length;
+					if (response.assets[assetLength - 1].params) {
+						console.log(
+							"Gov Token Info:",
+							response.assets[assetLength - 1].params
+						);
+					}
+				} else {
+					console.warn("gov-token not found");
```
ratik21, line 67: just to confirm, from a UI perspective -- we need to "display" this warning somehow. Right?

ratik21 (approved):

src/components/WalletConnect.vue
rmeena840: Yes Right

src/components/WalletConnect.vue
rmeena840: ok moved in constant file

src/components/WalletConnect.vue
rmeena840: I have a question here. I tried below import 
`import { AlgoSigner } from "@algo-builder/web/build/algo-signer-types"`

but while trying to use AlgoSigner.connect() method I am getting an error that connect method is not defined. I tried checking other projects they were also using `any` for AlgoSigner. Is there another data type for AlgoSigner which I am missing?

ratik21: a) The import is wrong. You shouldn't import from build. Do `import { types } from "@algo-builder/web"`. Then use `types.AlgoSigner`

b) It throws an error because `connect()` method is not implemented in [this](https://github.com/scale-it/algo-builder/blob/master/packages/web/src/algo-signer-types.ts#L147) type definition. It should be. 

If you want, you can do it as a separate task.

rmeena840: Seems like `types.AlgoSigner` also does not exist. Getting error AlgoSigner type does not exist.

I made below code change:
`import { types } from "@algo-builder/web";
declare var AlgoSigner: types.AlgoSigner;`

I am using `"@algo-builder/web": "^3.2.0",` as package dependency. Is there any specific version in which this is available?

robert-zaremba: @rmeena840 see the develop branch. You can use github commit as a dependency version, instead of the release. In that case you should use the same version in all: web, runtime and algob in package.json

robert-zaremba: Also, I agree with @ratik21 - we want to limit to minimum the eslint disable. You should always justify that addition.

rmeena840: I checked the documentation and their documentation itself uses `any` as a datatype.
https://developer.algorand.org/solutions/wagering-dapp-with-a-serverless-vue-algosigner-frontend/#algosigner-api

I think we have to go with `any` here.

rmeena840: So it exists in window object. See below attached image.
Screenshot 2022-03-31 at 7 25 50 PM

PR #28, opened 2022-04-12
Added all dao screen

Megha-Dev-19:
- Added all DAO screen with search dao app with app ID
- UI integration of govt-token ASA ID
- Added locked and available tokens by the user by query user holding 

image
image

- Decoded global state of app 
image

- Error handling
image

### DIFF
--- a/src/components/AppHeader.vue
+++ b/src/components/AppHeader.vue
@@ -92,24 +92,28 @@ import DaoStore from "../store/DaoID";
 import WalletStore from "../store/WalletStore";
 import { searchForAccount, searchForApplication } from "@/indexer";
 import { storeToRefs } from "pinia";
+import {
+	GLOBAL_STATE_MAP_KEY,
+	LOCAL_STATE_MAP_KEY,
+} from "@/constants/constant";
 
 export default defineComponent({
 	components: {
 		WalletConnect,
 	},
 	data() {
-		const Daostore = storeToRefs(DaoStore());
+		const daoStore = storeToRefs(DaoStore());
 		return {
 			currentPageKey: 0,
 			NavigationKey: NavigationKey,
 			EndPoint,
-			dao_id: Daostore.dao_id,
-			govt_id: Daostore.govt_id,
-			name: Daostore.name,
-			availableTokens: Daostore.available,
-			lockedTokens: Daostore.locked,
+			dao_id: daoStore.dao_id,
+			govt_id: daoStore.govt_id,
+			name: daoStore.name,
+			availableTokens: daoStore.available,
+			lockedTokens: daoStore.locked,
 			showIDTextField: false,
-			global_app_state: Daostore.global_app_state,
+			global_app_state: daoStore.global_app_state,
 		};
 	},
 	methods: {
@@ -134,7 +138,9 @@ export default defineComponent({
 					.then((response) => {
 						if (response) {
 							this.global_app_state = response;
-							this.name = this.global_app_state.get("dao_name") as string;
+							this.name = this.global_app_state.get(
+								GLOBAL_STATE_MAP_KEY.DaoName
+							) as string;
 						}
 					})
 					.catch((error) => console.log(error));
@@ -144,7 +150,7 @@ export default defineComponent({
 						.then((response) => {
 							if (response.localStateMap) {
 								this.lockedTokens = response.localStateMap.get(
-									"deposit"
+									LOCAL_STATE_MAP_KEY.Deposit
 								) as number;
 								this.availableTokens =
 									response.total_amount - this.lockedTokens;
--- a/src/components/AppHeader.vue
+++ b/src/components/AppHeader.vue
@@ -89,13 +89,8 @@ import { defineComponent } from "vue";
 import WalletConnect from "./WalletConnect.vue";
 import { NavigationKey, EndPoint } from "../types/enum.types";
 import DaoStore from "../store/DaoID";
-import WalletStore from "../store/WalletStore";
-import { searchForAccount, searchForApplication } from "@/indexer";
+import { searchApplicationAndAccount } from "@/indexer";
 import { storeToRefs } from "pinia";
-import {
-	GLOBAL_STATE_MAP_KEY,
-	LOCAL_STATE_MAP_KEY,
-} from "@/constants/constant";
 
 export default defineComponent({
 	components: {
@@ -130,36 +125,11 @@ export default defineComponent({
 		handleIDListener() {
 			this.showIDTextField = true;
 		},
-		searchID() {
+		async searchID() {
 			this.showIDTextField = false;
 			if (this.dao_id) {
 				this.dao_id = +this.dao_id;
-				searchForApplication(+this.dao_id)
-					.then((response) => {
-						if (response) {
-							this.global_app_state = response;
-							this.name = this.global_app_state.get(
-								GLOBAL_STATE_MAP_KEY.DaoName
-							) as string;
-						}
-					})
-					.catch((error) => console.log(error));
-				const walletStore = WalletStore();
-				if (walletStore.address) {
-					searchForAccount(walletStore.address, +this.dao_id)
-						.then((response) => {
-							if (response.localStateMap) {
-								this.lockedTokens = response.localStateMap.get(
-									LOCAL_STATE_MAP_KEY.Deposit
-								) as number;
-								this.availableTokens =
-									response.total_amount - this.lockedTokens;
-							}
-						})
-						.catch((error) => {
-							console.log(error);
-						});
-				}
+				searchApplicationAndAccount().catch((error) => console.log(error));
 			}
 		},
 	},
--- a/src/components/AppHeader.vue
+++ b/src/components/AppHeader.vue
@@ -89,7 +89,8 @@ import { defineComponent } from "vue";
 import WalletConnect from "./WalletConnect.vue";
 import { NavigationKey, EndPoint } from "../types/enum.types";
 import DaoStore from "../store/DaoID";
-import { searchForApplication } from "@/indexer";
+import WalletStore from "../store/WalletStore";
+import { searchForAccount, searchForApplication } from "@/indexer";
 import { storeToRefs } from "pinia";
 
 export default defineComponent({
@@ -108,6 +109,7 @@ export default defineComponent({
 			availableTokens: Daostore.available,
 			lockedTokens: Daostore.locked,
 			showIDTextField: false,
+			global_app_state: Daostore.global_app_state,
 		};
 	},
 	methods: {
@@ -127,7 +129,30 @@ export default defineComponent({
 		searchID() {
 			this.showIDTextField = false;
 			if (this.dao_id) {
-				searchForApplication(+this.dao_id);
+				searchForApplication(+this.dao_id)
+					.then((response) => {
+						if (response) {
+							this.global_app_state = response;
+							this.name = this.global_app_state.get("dao_name") as string;
+						}
+					})
+					.catch((error) => console.log(error));
+				const walletStore = WalletStore();
+				if (walletStore.address) {
+					searchForAccount(walletStore.address, +this.dao_id)
+						.then((response) => {
+							if (response.localStateMap) {
+								this.lockedTokens = response.localStateMap.get(
+									"deposit"
+								) as number;
+								this.availableTokens =
+									response.total_amount - this.lockedTokens;
+							}
+						})
+						.catch((error) => {
+							console.log(error);
+						});
+				}
 			}
 		},
 	},
--- a/src/components/AppHeader.vue
+++ b/src/components/AppHeader.vue
@@ -19,12 +19,14 @@
 								>Create DAO</a-button
 							>
 
-							<a-button
-								class="menu_option"
-								:type="isLinkActive(NavigationKey.DAOs)"
-								@click="() => handleMenuClick(NavigationKey.DAOs)"
-								>All DAOs</a-button
-							>
+							<router-link :to="{ path: EndPoint.ALL_DAO }">
+								<a-button
+									class="menu_option"
+									:type="isLinkActive(NavigationKey.DAOs)"
+									@click="() => handleMenuClick(NavigationKey.DAOs)"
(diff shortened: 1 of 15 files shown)


src/pages/AllDao.vue
```diff
@@ -0,0 +1,65 @@
...
+		const onFinishFailed = (errorinfo: Event) => {
+			console.warn("Failed:", errorinfo);
+		};
+
+		const validateMessages = {
+			required: "required!",
+			types: {
+				number: "It is is not a valid number!",
+			},
+		};
```
rmeena840: Move all required methods into the method block. In this way, it will be available to the entire component without returning all the methods.

src/indexer/index.ts
```diff
@@ -11,3 +17,44 @@ export const searchForAssets = async (tokenName: string) => {
...
+};
+
+export async function readAppGlobalState(
+	creator: AccountAddress,
+	appID: number
+): Promise<Map<Key, StateValue> | undefined> {
+	const accountInfoResponse = await algodClient
+		.accountInformation(creator)
+		.do();
+	for (const app of accountInfoResponse["created-apps"]) {
```
rmeena840: You can also use the array filter method. No need to iterate all values.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

rmeena840 (changes requested):
Left few comments.

src/indexer/index.ts
Megha-Dev-19: actually this is a function from algobuilder. I updated it here to be used in web. Will update it after consulting with @robert-zaremba

robert-zaremba: both options are fine. We are iterating a map here.

src/components/AppHeader.vue
```diff
@@ -89,13 +123,39 @@ export default defineComponent({
...
-		return {
-			id,
-		};
+		handleIDListener() {
+			this.showIDTextField = true;
+		},
+		searchID() {
+			this.showIDTextField = false;
+			if (this.dao_id) {
+				this.dao_id = +this.dao_id;
```
vuvoth: What is `+this.dao_id` mean?

src/indexer/index.ts
```diff
@@ -1,3 +1,9 @@
+import algodClient from "@/config/algob.config";
+import {
+	Key,
+	StateValue,
+	AccountAddress,
+} from "@algo-builder/algob/build/types";
```
vuvoth, line 14: I don't think import from `build/types` is good approach.

src/pages/AllDao.vue
```diff
@@ -0,0 +1,103 @@
...
+
+export default defineComponent({
+	name: "AllDao",
+	data() {
+		return {
+			error: "",
+		};
+	},
+	methods: {
+		onFinish(values: any) {
```
vuvoth, line 50: We should avoid use `any` type.

src/pages/Proposal.vue
```diff
@@ -142,60 +143,44 @@ export default defineComponent({
...
-		};
-		const onFinishFailed = (errorinfo: Event) => {
+		},
+		onFinishFailed(errorinfo: Event) {
 			console.warn("Failed:", errorinfo);
-		};
-		// eslint-disable-next-line
-		const disabledDate = (current: any) => {
+		},
+		disabledDate(current: any) {
```
vuvoth: we shouldn't use any type.

src/types/interface.types.ts
```diff
@@ -1,14 +1,20 @@
...
 export interface WalletStoreState {
 	walletKind: WalletType;
 	webMode: WebMode;
 	address: string;
 }
 
-export interface DaoID {
-	id: number | undefined;
+export interface DAO {
+	dao_id: number | undefined;
```
vuvoth: How about use syntax `dao_id?: number`? Choice the best way as you refer <3

src/pages/Proposal.vue
```diff
@@ -142,60 +143,44 @@ export default defineComponent({
...
 				current < new Date(new Date().valueOf() - 1000 * 60 * 60 * 24)
 			);
-		};
-
-		const range = (start: number, end: number) => {
+		},
+		range(start: number, end: number) {
 			const result = [];
 
 			for (let i = start; i < end; i++) {
```
vuvoth, line 164: We can use range function at here, right?

src/pages/Proposal.vue
```diff
@@ -142,60 +143,44 @@ export default defineComponent({
...
 			console.warn("Failed:", errorinfo);
-		};
-		// eslint-disable-next-line
-		const disabledDate = (current: any) => {
+		},
+		disabledDate(current: any) {
 			// Can not select day before today
 			return (
 				current &&
 				current < new Date(new Date().valueOf() - 1000 * 60 * 60 * 24)
```
vuvoth: Let create constant to store `1000 * 60 * 60 * 24`

src/pages/AllDao.vue
```diff
@@ -0,0 +1,103 @@
...
+							) as number;
+							this.formState.available =
+								response.total_amount - this.formState.locked;
+						}
+					})
+					.catch((error) => {
+						this.error = error.message;
+						setTimeout(() => {
+							this.error = "";
+						}, 2000);
```
vuvoth: Why we need `setTimeout` here?

src/indexer/index.ts
```diff
@@ -11,3 +17,87 @@ export const searchForAssets = async (tokenName: string) => {
...
+export async function readAppLocalState(
+	account: AccountAddress,
+	appID: number
+): Promise<Map<Key, StateValue> | undefined> {
+	const accountInfoResponse = await algodClient
+		.accountInformation(account)
+		.do();
+	for (const app of accountInfoResponse["apps-local-state"]) {
+		if (app.id === appID) {
+			const localStateMap = new Map<Key, StateValue>();
```
vuvoth: I see we share same logic when create {local,global}StateMap. Could we create function for it?

src/indexer/index.ts
```diff
@@ -11,3 +17,87 @@ export const searchForAssets = async (tokenName: string) => {
...
+		const creator = applicationInfo.application.params.creator;
+		const globalState = await readAppGlobalState(creator, application_id);
+		return globalState;
+	} catch (e) {
+		console.log(e);
+		throw e;
+	}
+};
+
+export const searchForAccount = async (
```
vuvoth, line 54: What is type of this function?

src/indexer/index.ts
```diff
@@ -11,3 +17,87 @@ export const searchForAssets = async (tokenName: string) => {
 		console.log(e);
 	}
 };
+
+export const searchForApplication = async (application_id: number) => {
```
vuvoth: What is type of this function?

src/config/algob.config.ts
```diff
@@ -0,0 +1,10 @@
+import algosdk from "algosdk";
```
vuvoth, line 1: Why we need algob.config.ts?

vuvoth (commented):
Left few comments! Thanks for your PR!

src/components/AppHeader.vue
Megha-Dev-19: because it saved number as a string `3 => "3"` so to make `3` I use +
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus

src/indexer/index.ts
Megha-Dev-19: okay, I will create a type here itself

src/pages/AllDao.vue
Megha-Dev-19: yes I know but antd form doesn't have any specific type mentioned in their docs, they also use any, that's why using any

src/pages/AllDao.vue
Megha-Dev-19: to remove the error from screen, it displays the error for 2 seconds and then shows the form again

src/config/algob.config.ts
Megha-Dev-19: because webmode currently doesn't support query localstate and globalstate so to get those values I use algodclient

src/indexer/index.ts
Megha-Dev-19: yes that's right just the keys are different, will update

src/components/AppHeader.vue
```diff
@@ -65,17 +88,28 @@
...
+import WalletStore from "../store/WalletStore";
+import { searchForAccount, searchForApplication } from "@/indexer";
+import { storeToRefs } from "pinia";
 
 export default defineComponent({
 	components: {
 		WalletConnect,
 	},
 	data() {
+		const Daostore = storeToRefs(DaoStore());
```
rmeena840: Naming convention. It should be `daoStore`

src/pages/AllDao.vue
```diff
@@ -0,0 +1,103 @@
...
+	},
+	methods: {
+		onFinish(values: any) {
+			searchForApplication(+values.dao_id)
+				.then((response) => {
+					console.log("Global App Info:", response);
+					if (response) {
+						this.formState.global_app_state = response;
+						this.formState.name = this.formState.global_app_state.get(
+							"dao_name"
```
rmeena840: There are many such constant strings available. We should move them all to a constant file. Same local_app_state or wherever applicable.

src/pages/AllDao.vue
```diff
@@ -0,0 +1,103 @@
...
+						this.error = "";
+					}, 1500);
+				});
+			const walletStore = WalletStore();
+			if (walletStore.address) {
+				searchForAccount(walletStore.address, +values.dao_id)
+					.then((response) => {
+						if (response.localStateMap) {
+							this.formState.locked = response.localStateMap.get(
+								"deposit"
```
rmeena840: ditto
constant string....

src/pages/Proposal.vue
```diff
@@ -142,60 +143,44 @@ export default defineComponent({
...
 			for (let i = start; i < end; i++) {
 				result.push(i);
 			}
 
 			return result;
-		};
+		},
 		// eslint-disable-next-line
-		const disabledRangeTime = (_: any, type: "start" | "end") => {
+		disabledRangeTime(_: any, type: "start" | "end") {
```
rmeena840: We can move "start | end" to type

rmeena840 (changes requested):

src/types/interface.types.ts
Megha-Dev-19: oh yes, it completely slipped my mind, will update

src/pages/Proposal.vue
Megha-Dev-19: we don't have a direct range function in javascript, rather we use for loop

vuvoth: Ah,  I see it.

src/config/algob.config.ts
vuvoth: I think we don't need it. The network user use on `Algosigner` and in `algob.config.ts` can be **different** and this is big problem.

We can use [algosigner.algod](https://github.com/PureStake/algosigner/blob/develop/docs/dApp-integration.md#algosigneralgod-ledger-mainnettestnet-path-algod-v2-path--) in this case and this method ensure we use the same client to query data and send txn to network.
 
You can query globalstates and localstates of application by the endpoint `GET /v2/applications/{application-id}`. More details at [here](https://developer.algorand.org/docs/rest-apis/algod/v2/#get-v2applicationsapplication-id)

Megha-Dev-19: thanks for the advice, I didn't thought that way

src/components/AppHeader.vue
```diff
@@ -46,13 +48,34 @@
...
+										<div v-else>Enter ID</div>
+									</div>
+								</div></a-descriptions-item
+							>
+							<a-descriptions-item label="DAO Name">{{
+								name
+							}}</a-descriptions-item>
 							<a-descriptions-item label="Govt Token ID">
-								421421
+								{{ govt_id }}
```
vuvoth: Let use camelCase  style

src/components/AppHeader.vue
```diff
@@ -65,17 +88,27 @@
...
 	components: {
 		WalletConnect,
 	},
 	data() {
+		const daoStore = storeToRefs(DaoStore());
 		return {
 			currentPageKey: 0,
 			NavigationKey: NavigationKey,
 			EndPoint,
+			dao_id: daoStore.dao_id,
```
vuvoth: Same here

src/indexer/index.ts
```diff
@@ -1,13 +1,147 @@
...
+	LOCAL_STATE_MAP_KEY,
+} from "@/constants/constant";
+import DaoID from "@/store/DaoID";
+import WalletStore from "@/store/WalletStore";
+import {
+	Key,
+	StateValue,
+	AccountAddress,
+} from "@algo-builder/algob/build/types";
+declare var AlgoSigner: any; // eslint-disable-line
```
vuvoth, line 15: We defined algosigner type at [here](https://github.com/scale-it/algo-builder/blob/master/packages/web/src/algo-signer-types.ts). Could we use it?

src/indexer/index.ts
```diff
@@ -1,13 +1,147 @@
...
+		const creator = applicationInfo.application.params.creator;
+		const globalStateMap = await readAppState(
+			GLOBAL_STATE,
+			creator,
+			application_id
+		);
+		return globalStateMap;
+	} catch (e) {
+		console.log(e);
+		throw e;
```
vuvoth, line 50: Why you throw error at here but not throw error in `searchForAssets`

vuvoth (commented):
LGTM! Left few comments!

src/indexer/index.ts
Megha-Dev-19: yes, that's the way of accessing algosigner extension

src/indexer/index.ts
Megha-Dev-19: I guess I missed it, will update

vuvoth (approved):

rmeena840 (approved):

PR #30, opened 2022-04-27
Vote, Deposit Tokens, Withdraw Tokens page added

Megha-Dev-19:
- Added vote token page with deposit and withdrawal tabs
image

image

- Added vote page
image

- Added notification banner in case of success
- Added error close button

### DIFF
--- a/src/constants/error.constant.ts
+++ b/src/constants/error.constant.ts
@@ -1,5 +1,6 @@
 import DaoID from "@/store/DaoID";
 import WalletStore from "@/store/WalletStore";
+import { message, notification } from "ant-design-vue";
 import { reactive } from "vue";
 
 export const APP_NOT_FOUND = "Please add DAO App ID";
@@ -20,4 +21,20 @@ export const OverallErrorCheck = () => {
 		return WALLET_NOT_CONNECT;
 	}
 	return "";
+};
+
+export const openErrorNotificationWithIcon = (
+	message: string,
+	description: string
+) => {
+	notification["error"]({
+		message: message,
+		description: description,
+		duration: 0, // to close only when user prompts close button
+		onClose: () => notification.close(""),
+	});
+};
+
+export const ErrorMessage = (key: string) => {
+	message.error({ content: "Error Occured", key: key });
 };
--- a/src/constants/error.constant.ts
+++ b/src/constants/error.constant.ts
@@ -7,7 +7,7 @@ export const APP_NOT_FOUND = "Please add DAO App ID";
 export const TOKEN_NOT_FOUND = "Govt token not found";
 export const WALLET_NOT_CONNECT = "Please connect to your Wallet";
 
-export const OverallErrorCheck = () => {
+export const overallErrorCheck = () => {
 	const daoIDStore = reactive(DaoID());
 	const walletStore = reactive(WalletStore());
 
@@ -35,6 +35,6 @@ export const openErrorNotificationWithIcon = (
 	});
 };
 
-export const ErrorMessage = (key: string) => {
+export const errorMessage = (key: string) => {
 	message.error({ content: "Error Occured", key: key });
 };
--- a/src/constants/error.constant.ts
+++ b/src/constants/error.constant.ts
@@ -1,0 +1,3 @@
+export const APP_NOT_FOUND = "Please add DAO App ID";
+export const TOKEN_NOT_FOUND = "Govt token not found";
+export const WALLET_NOT_CONNECT = "Please connect to your Wallet";
--- a/src/constants/error.constant.ts
+++ b/src/constants/error.constant.ts
@@ -1,3 +1,23 @@
+import DaoID from "@/store/DaoID";
+import WalletStore from "@/store/WalletStore";
+import { reactive } from "vue";
+
 export const APP_NOT_FOUND = "Please add DAO App ID";
 export const TOKEN_NOT_FOUND = "Govt token not found";
 export const WALLET_NOT_CONNECT = "Please connect to your Wallet";
+
+export const OverallErrorCheck = () => {
+	const daoIDStore = reactive(DaoID());
+	const walletStore = reactive(WalletStore());
+
+	if (typeof daoIDStore.dao_id === "undefined") {
+		return APP_NOT_FOUND;
+	}
+	if (typeof daoIDStore.govt_id === "undefined") {
+		return TOKEN_NOT_FOUND;
+	}
+	if (!walletStore.address.length) {
+		return WALLET_NOT_CONNECT;
+	}
+	return "";
+};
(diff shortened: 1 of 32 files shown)


src/pages/Vote.vue
```diff
@@ -40,33 +40,90 @@
...
+			const registerVoteParam: types.ExecParams = {
+				type: types.TransactionType.CallApp,
+				sign: types.SignType.SecretKey,
+				fromAccount: {
+					addr: this.walletStore.address,
+					sk: new Uint8Array(0),
+				},
+				appID: this.daoIDStore.dao_id,
+				payFlags: { totalFee: 2000 },
+				appArgs: [DAOActions.REGISTER_VOTE, `str:${this.formState.vote_type}`],
```
vuvoth: When I run this branch in my laptop, this.formState.vote_type return number type not string. It should be VoteOptions type(yes,no,abstain)

Please check it!

src/indexer/index.ts
```diff
@@ -146,32 +146,34 @@ export const searchApplicationAndAccount = async () => {
...
 		proposalStore.setProposalAddr(lsig.address());
-
+		searchForGovASAToken();
 		if (account && account.localStateMap) {
-			daoIdStore.locked = account.localStateMap.get(
-				LOCAL_STATE_MAP_KEY.Deposit
-			) as number;
-			daoIdStore.available = account.total_amount - daoIdStore.locked;
+			daoIdStore.locked =
+				(account.localStateMap.get(LOCAL_STATE_MAP_KEY.Deposit) as number) ?? 0;
```
rmeena840, line 152: I think this line is sufficient. `daoIdStore.locked` value will be set anyhow even `localStateMap` is present or undefined.
Just suggestion but please verify once.

src/indexer/index.ts
```diff
@@ -181,25 +183,23 @@ export const isAssetOpted = async (address: string, asset_id: number) => {
...
-			parsedOptedApplicationInfo &&
-			parsedOptedApplicationInfo["apps-local-states"]
-		) {
-			const isGivenApplicationOpted = parsedOptedApplicationInfo[
-				"apps-local-states"
-			].find((element: any) => element["id"] === application_id);
+		if (optedApplicationInfo && optedApplicationInfo["app-local-state"]) {
+			const isGivenApplicationOpted =
+				optedApplicationInfo["app-local-state"].id === application_id;
 			if (isGivenApplicationOpted) return true;
```
vuvoth: We only need to return `optedApplicationInfo["app-local-state"].id === application_id;`

src/indexer/index.ts
```diff
@@ -146,32 +146,34 @@ export const searchApplicationAndAccount = async () => {
...
-			daoIdStore.locked = account.localStateMap.get(
-				LOCAL_STATE_MAP_KEY.Deposit
-			) as number;
-			daoIdStore.available = account.total_amount - daoIdStore.locked;
+			daoIdStore.locked =
+				(account.localStateMap.get(LOCAL_STATE_MAP_KEY.Deposit) as number) ?? 0;
+		} else {
+			daoIdStore.locked = 0;
 		}
+		daoIdStore.available = account.total_amount - daoIdStore.locked;
```
rmeena840: Can total_amount be undefined? It might give runtime error if we get undefined error.

src/pages/Proposal.vue
```diff
@@ -268,27 +270,17 @@ export default defineComponent({
 					},
 				];
 				let response = await this.walletStore.webMode.executeTx(addProposalTx);
+				openSuccessNotificationWithIcon(
+					"Success",
+					"Your Proposal has benn created."
```
rmeena840: Please move msg to new file inside success.constant.ts

src/pages/Proposal.vue
```diff
@@ -181,16 +189,10 @@ export default defineComponent({
 				} = values;
 				if (typeof this.daoStore.dao_id === "undefined") {
```
vuvoth: We duplicated code when check error in Proposal and Vote
https://github.com/scale-it/sigmadao-ui/blob/ac184aaaeb2fb55e7b8489413405fcc7104eb629/src/pages/DepositToken.vue#L77

src/pages/Vote.vue
```diff
@@ -39,34 +46,86 @@
...
 		return {
 			formState,
+			daoIDStore,
+			walletStore,
 			validateMessages: VALIDATE_MESSAGES,
 		};
 	},
+	methods: {
+		async onFinish() {
+			if (typeof this.daoIDStore.dao_id === "undefined") {
```
vuvoth: Also duplicated code at here.
Could we remove duplicated code?

vuvoth (commented):
Nice work! I left few comments.

rmeena840 (changes requested):

src/indexer/index.ts
Megha-Dev-19: oh, yes that will do it, will update

src/pages/Proposal.vue
Megha-Dev-19: yes, error conditions are repeated at multiple places, will create a common function and return error from there

src/indexer/index.ts
Megha-Dev-19: But we are setting the `locked` value from `localStateMap` as it has a deposit key, so how will it be updated if it is undefined..?  
May be I m not able to understand what you mean. Can you please elaborate.?

src/indexer/index.ts
Megha-Dev-19: Actually I have to remove this, `total_amount` is algos but we needed to show gov-token so I updated the UI but forgot to remove this line, will update

src/pages/Proposal.vue
Megha-Dev-19: okay sure

vuvoth (approved):
LGTM!

src/constants/error.constant.ts
```diff
@@ -1,2 +1,40 @@
+import DaoID from "@/store/DaoID";
+import WalletStore from "@/store/WalletStore";
+import { message, notification } from "ant-design-vue";
+import { reactive } from "vue";
+
 export const APP_NOT_FOUND = "Please add DAO App ID";
 export const TOKEN_NOT_FOUND = "Govt token not found";
+export const WALLET_NOT_CONNECT = "Please connect to your Wallet";
+
+export const OverallErrorCheck = () => {
```
rmeena840: naming convention of method name.

rmeena840 (approved):
Left one comment. Pre-approving.

src/constants/error.constant.ts
Megha-Dev-19: updated

PR #31, opened 2022-05-04
use local indexer for SigmaDAO

rmeena840:
- use local indexer for SigmaDAO
- Make SigmaDao compatible with latest algosdk

### DIFF
--- a/src/indexer/index.ts
+++ b/src/indexer/index.ts
@@ -15,16 +15,17 @@ import {
 } from "@algo-builder/algob/build/types";
 import type { LogicSigAccount } from "algosdk";
 import { getProposalLsig } from "../contract/dao";
+import indexerClient from "../config/indexer.config";
 declare var AlgoSigner: any; // eslint-disable-line
 
 export const searchForAssets = async (
 	tokenName: string
 ): Promise<Record<string, any> | undefined> => {
 	try {
-		const accountInfo = await AlgoSigner.indexer({
-			ledger: CHAIN_NAME,
-			path: `/v2/assets?name=${tokenName}`,
-		});
+		const accountInfo = await indexerClient
+			.searchForAssets()
+			.name(tokenName)
+			.do();
 		return JSON.parse(JSON.stringify(accountInfo));
 	} catch (e) {
 		console.log(e);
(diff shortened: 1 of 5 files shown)


vuvoth (commented):
Where is file `/config/indexer.config`?

rmeena840:
@vuvoth It's there inside config folder.

rmeena840:
@Megha-Dev-19  Please review the PR.

src/indexer/index.ts
```diff
@@ -15,16 +15,17 @@ import {
 } from "@algo-builder/algob/build/types";
 import type { LogicSigAccount } from "algosdk";
 import { getProposalLsig } from "../contract/dao";
+import indexerClient from "../config/indexer.config";
```
vuvoth, line 18: Same like `config/indexer.config` only work with local network. 
I recommend we import those value from env.
Example: 
```javascript
const indexer_token = process.env.INDEXER_TOKEN || "";
const indexer_server = process.env.INDEXER_URL || "http://localhost";
const indexer_port = process.env.INDEXER_PORT || 8980;
```

src/indexer/index.ts
```diff
@@ -15,16 +15,17 @@ import {
...
 
 export const searchForAssets = async (
 	tokenName: string
 ): Promise<Record<string, any> | undefined> => {
 	try {
-		const accountInfo = await AlgoSigner.indexer({
-			ledger: CHAIN_NAME,
-			path: `/v2/assets?name=${tokenName}`,
-		});
+		const accountInfo = await indexerClient
```
vuvoth, line 25: I think we should create Indexer instance in src/indexer. How you think about it? I not sure this is good idea.

vuvoth (commented):
Thank for your PR! I left a comments.

src/indexer/index.ts
rmeena840: Yes, The instance is created inside `config/indexer.config`. I think the current path is good because this also includes ports and URL which needed for instance.

src/indexer/index.ts
rmeena840: Added

Megha-Dev-19 (approved):
LGTM

vuvoth (approved):
Nice work!

PR #32, opened 2022-05-05
Network switch and multiple wallet support added

Megha-Dev-19:
- Added multiple wallet and network switch support

### Note : 
1. Private net is only for testing purpose, also wallet connect and myalgo wallet doesn't support it 
2. Changed `webMode` type to `any` in `fundAmount` since all wallets mode have different type after connection which aren't specified in web package i.e it is different from `WebModeTypes`
3. I have not updated the indexer network since we will upgrade to local indexer.

image

### DIFF
--- a/src/utility/algod.utility.ts
+++ b/src/utility/algod.utility.ts
@@ -1,4 +1,4 @@
-import { types, tx as webTx, mkTxParams, WebMode } from "@algo-builder/web";
+import { types, tx as webTx, mkTxParams } from "@algo-builder/web";
 import { types as aTypes } from "@algo-builder/algob";
 import algodClient from "@/config/algob.config";
 import type { LogicSigAccount } from "algosdk";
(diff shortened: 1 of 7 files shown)


src/utility/algod.utility.ts
```diff
@@ -11,9 +11,9 @@ export const fundAmount = async (
 	from: string,
 	to: string,
 	amount: number,
-	webMode: WebMode
+	webMode: any // eslint-disable-line
```
rmeena840, line 14: Can't we make it type `WebModeTypes`?

rmeena840 (approved):
Good work!! LGTM. Left one comment. Pre-approving.

rmeena840:
@Megha-Dev-19 You can now delete this file `algosigner.config`. We are getting NetworkType from enum now.

src/utility/algod.utility.ts
Megha-Dev-19: No we can’t because after connecting the wallet new type is assigned to my algo connect and wallet connect

Megha-Dev-19:
> @Megha-Dev-19 You can now delete this file `algosigner.config`. We are getting NetworkType from enum now.

Yeah sure, your local indexer PR wasn’t merged  so I didn’t do it

PR #33, opened 2022-05-06
Account switch feature in SigmaDao

rmeena840:
- Account switch feature in SigmaDao
- Latest Antd version support added

Screenshot 2022-05-09 at 1 54 13 PM


robert-zaremba:
This doesn't look good. The "selected address" has additional frame and "selected wallet" doesn't have it.
I would put the label outside the frame, so only the address will be in the select component.

robert-zaremba:
Also, can you make sure this UI also works on Mobile?
I suggest that we don't show the full address... instead let's only show 4 first characters and last 4 characters, eg:
```
HV51...2AOU
```

src/components/WalletConnect.vue
```diff
@@ -52,23 +52,36 @@
...
+						<a-dropdown>
+							<template #overlay>
+								<a-menu @click="handleAccountSwitch">
+									<a-menu-item v-for="addr in walletAddresses" :key="addr">
+										{{ addr }}
+									</a-menu-item>
+								</a-menu>
+							</template>
+							<a-button>
+								<span v-if="!walletAddresses.length">No Address Found</span>
```
Megha-Dev-19: we are already in v-else from `!walletaddress` template that means this won't be called ever.

Megha-Dev-19 (commented):
Can you check in myalgo wallet, I believe there we do have a option to select address from wallet itself, and for wallet connect we have to scan the QR so I believe there isn't any address array available? Can you confirm the same?

rmeena840:
> This doesn't look good. The "selected address" has additional frame and "selected wallet" doesn't have it. I would put the label outside the frame, so only the address will be in the select component.

Changes done. Please check the PR description for an updated screenshot.

rmeena840:
> Also, can you make sure this UI also works on Mobile? I suggest that we don't show the full address... instead let's only show 4 first characters and last 4 characters, eg:
> 
> ```
> HV51...2AOU
> ```

I have added this format. Please check the updated PR description.

src/components/WalletConnect.vue
rmeena840: No, we are not in v-else. First v-if will be checked where I have put check condition on wallet address length.

rmeena840:
> Can you check in myalgo wallet, I believe there we do have a option to select address from wallet itself, and for wallet connect we have to scan the QR so I believe there isn't any address array available? Can you confirm the same?

@Megha-Dev-19 Yes it returns the account address array: https://connect.myalgo.com/docs/getting-started/myalgo-connect-api/#response

rmeena840:
> Also, can you make sure this UI also works on Mobile? I suggest that we don't show the full address... instead let's only show 4 first characters and last 4 characters, eg:
> 
> ```
> HV51...2AOU
> ```

@robert-zaremba I have added this format. Please check the updated PR description.

rmeena840:
> Changes done. Please check the PR description for an updated screenshot.

@robert-zaremba Changes done. Please check the PR description for an updated screenshot.

src/components/WalletConnect.vue
Megha-Dev-19: image

The topmost template has the v-if for `walletaddress` length, thus the second template won't show if `walletaddress` is empty, also I have tested all wallets and I have seen it doesn't allow you to grant access unless there is atleast one account available in the network, so I think we don't need the check

Megha-Dev-19:
> @Megha-Dev-19 Yes it returns the account address array: https://connect.myalgo.com/docs/getting-started/myalgo-connect-api/#response

That's not what I meant, here in my algo wallet we can select the account from the wallet itself, so I believe its better to allow wallet to handle account switch, we can have account switch with algosigner only, since wallet connect doesn't return accounts array (please confirm). Also see if we need to make any changes in myalgo wallet for this account switch functionality

image

src/components/WalletConnect.vue
```diff
@@ -52,23 +52,45 @@
...
+							<span style="margin-top: 5px">
+								<a-typography-text strong>Wallet: </a-typography-text>
+								<a-typography-link>{{ selectedWallet }}</a-typography-link>
+							</span>
+						</a-col>
+						<a-col>
+							<a-dropdown>
+								<template #overlay>
+									<a-menu @click="handleAccountSwitch">
+										<a-menu-item v-for="addr in walletAddresses" :key="addr">
```
vuvoth, line 75: What happened if `walletAddresses` is empty? Should we handle this case

src/components/WalletConnect.vue
rmeena840: Ok got it.

src/components/WalletConnect.vue
rmeena840: Already handled.

src/components/WalletConnect.vue
```diff
@@ -246,6 +272,14 @@ export default defineComponent({
 				this.walletStore.setNetworkTypes(e.key);
 			}
 		},
+		handleAccountSwitch(e: any) {
+			if (e.key) {
+				const addr = e.key;
+				this.walletAddress = addr;
+				this.setAddress(addr);
+				console.log("Account Switched.");
```
robert-zaremba: let's remove this log, or add address to the log.

src/styles/wallet.css
```diff
@@ -1,5 +1,5 @@
 .wallet {
-	width: 60%;
+	width: 66%;
```
robert-zaremba, line 2: I think this is too big. How it will look like in smaller screens?

robert-zaremba (commented):
- let's address @Megha-Dev-19  comments
- let's make sure it will look good on mobile.

I will let Megha to do the final approval.

src/styles/wallet.css
rmeena840: I have shared with you the mobile screen version

rmeena840: Since we have added a dropdown for the address so space is needed. So we have to increase it.

src/components/WalletConnect.vue
rmeena840: Done

rmeena840:
> > @Megha-Dev-19 Yes it returns the account address array: https://connect.myalgo.com/docs/getting-started/myalgo-connect-api/#response
> 
> That's not what I meant, here in my algo wallet we can select the account from the wallet itself, so I believe its better to allow wallet to handle account switch, we can have account switch with algosigner only, since wallet connect doesn't return accounts array (please confirm). Also see if we need to make any changes in myalgo wallet for this account switch functionality
> 
> image

We discussed and changes done.

rmeena840:
> > > @Megha-Dev-19 Yes it returns the account address array: https://connect.myalgo.com/docs/getting-started/myalgo-connect-api/#response
> > 
> > 
> > That's not what I meant, here in my algo wallet we can select the account from the wallet itself, so I believe its better to allow wallet to handle account switch, we can have account switch with algosigner only, since wallet connect doesn't return accounts array (please confirm). Also see if we need to make any changes in myalgo wallet for this account switch functionality
> > image
> 

We discussed and changes done.

rmeena840:
@Megha-Dev-19  Discussed changes done. Please review again.

src/components/WalletConnect.vue
```diff
@@ -246,6 +270,14 @@ export default defineComponent({
...
 			}
 		},
+		handleAddressSwitch(e: any) {
+			if (e.key) {
+				const addr = e.key;
+				this.walletAddress = addr;
+				this.setAddress(addr);
+				console.log("Address Switched.");
+			}
+		},
```
Megha-Dev-19: When the user changes the address, please update the tokens information, i.e available and locked

Megha-Dev-19:
> @Megha-Dev-19 Discussed changes done. Please review again.

Thanks for updating.

src/components/WalletConnect.vue
rmeena840: Done

Megha-Dev-19 (approved):

PR #34, opened 2022-05-09
Opt in DAO app button 

Megha-Dev-19:
image

### DIFF
--- a/src/components/AppHeader.vue
+++ b/src/components/AppHeader.vue
@@ -105,11 +105,13 @@ import DaoStore from "../store/DaoID";
 import { isApplicationOpted, searchApplicationAndAccount } from "@/indexer";
 import { storeToRefs } from "pinia";
 import {
+	daoAppMessage,
 	errorMessage,
 	loadingMessage,
 	openErrorNotificationWithIcon,
 	openSuccessNotificationWithIcon,
 	successMessage,
+	UNSUCCESSFUL,
 } from "@/constants";
 import WalletStore from "@/store/WalletStore";
 import { optInDaoApp } from "@/utility";
@@ -134,7 +136,7 @@ export default defineComponent({
 			showIDTextField: false,
 			key: "HeaderKey",
 			walletStore,
-			removeDaoData: DaoStore().removeDaoID,
+			resetDaoStore: DaoStore().resetDaoStore,
 		};
 	},
 	methods: {
@@ -161,7 +163,7 @@ export default defineComponent({
 						successMessage(this.key);
 						openSuccessNotificationWithIcon(
 							"Successful",
-							`Your DAO App of ID ${this.daoID} is selected.`
+							daoAppMessage.SUCCESSFUL(this.daoID as number)
 						);
 						if (this.walletStore.address) {
 							isApplicationOpted(this.walletStore.address, this.daoID as number)
@@ -169,13 +171,13 @@ export default defineComponent({
 									this.showOptIn = !appIsOptedIn;
 									if (appIsOptedIn) {
 										openSuccessNotificationWithIcon(
-											"You have already Opted-in DAO App"
+											daoAppMessage.ALREADY_OPT_IN
 										);
 									}
 								})
 								.catch((error) =>
 									openErrorNotificationWithIcon(
-										"Unsuccessful while getting DAO App Opt-in Details",
+										daoAppMessage.UNSUCCESFUL,
 										error.message
 									)
 								);
@@ -183,11 +185,11 @@ export default defineComponent({
 					})
 					.catch((error) => {
 						errorMessage(this.key);
-						openErrorNotificationWithIcon("Unsuccessful", error.message);
+						openErrorNotificationWithIcon(UNSUCCESSFUL, error.message);
 					});
 			} else {
 				// when daoID is removed
-				this.removeDaoData();
+				this.resetDaoStore();
 			}
 		},
 		async optIn() {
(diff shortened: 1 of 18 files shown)


src/components/AppHeader.vue
```diff
@@ -152,13 +162,46 @@ export default defineComponent({
 							"Successful",
 							`Your DAO App of ID ${this.daoID} is selected.`
 						);
+						if (this.walletStore.address) {
+							isApplicationOpted(this.walletStore.address, this.daoID as number)
```
rmeena840, line 169: avoid `as number`. daoID can be undefined as well. Put a check for daoID undefined.

src/components/AppHeader.vue
```diff
@@ -152,13 +162,46 @@ export default defineComponent({
...
 					.catch((error) => {
 						errorMessage(this.key);
 						openErrorNotificationWithIcon("Unsuccessful", error.message);
 					});
 			}
 		},
+		async optIn() {
+			try {
+				await optInDaoApp(
+					this.walletStore.address,
```
rmeena840: same here.

src/indexer/index.ts
```diff
@@ -190,14 +190,17 @@ export const isApplicationOpted = async (
...
 ): Promise<boolean> => {
 	try {
-		const optedApplicationInfo = await AlgoSigner.algod({
-			ledger: CHAIN_NAME,
-			path: `/v2/accounts/${address}/applications/${application_id}`,
-		});
-		if (optedApplicationInfo && optedApplicationInfo["app-local-state"]) {
-			return optedApplicationInfo["app-local-state"].id === application_id;
+		let isAppOpted = false;
+		const applicationInfo = await indexerClient.lookupAccountByID(address).do();
```
rmeena840: Check here If using `indexerClient` gives same response as `AlgoSigner.algod`. I think it should be `algodClient`

src/pages/Proposal.vue
```diff
@@ -1,137 +1,142 @@
...
-					:rules="[{ required: true }]"
-				>
-					<a-range-picker
-						format="YYYY-MM-DD HH:mm:ss"
-						value-format="YYYY-MM-DD HH:mm:ss"
-						:disabled-date="disabledDate"
-						:disabled-time="disabledRangeTime"
-						showTime
-						v-model:value="formState.vote_date"
+	<opt-in-error>
```
rmeena840, line 2: Why are we changing Proposal.vue file for opt in button?

rmeena840 (changes requested):

src/components/AppHeader.vue
Megha-Dev-19: image
we are already testing it before

src/components/AppHeader.vue
Megha-Dev-19: we are showing optin button only when address and daoID is provided by user, see the condition when we set the `show_opt_in` state to true

src/indexer/index.ts
Megha-Dev-19: It gives the same response, its just that the endpoints are different, I updated it so as to minimise the use of algod for reading data since indexer is considered to be fast

src/pages/Proposal.vue
Megha-Dev-19: Robert asked me to not display any form until the user has opted in dao app, thus all three form are displayed only when app is opted in

src/components/AppHeader.vue
rmeena840: ok

src/indexer/index.ts
rmeena840: ok

src/pages/Proposal.vue
rmeena840: ok

src/components/AppHeader.vue
Megha-Dev-19: Added check

src/components/AppHeader.vue
```diff
@@ -152,11 +163,49 @@ export default defineComponent({
 							"Successful",
 							`Your DAO App of ID ${this.daoID} is selected.`
 						);
+						if (this.walletStore.address) {
+							isApplicationOpted(this.walletStore.address, this.daoID as number)
+								.then((appIsOptedIn: boolean) => {
+									this.showOptIn = !appIsOptedIn;
+									if (appIsOptedIn) {
+										openSuccessNotificationWithIcon(
+											"You have already Opted-in DAO App"
```
rmeena840: Move all messages in a constant file.

src/store/WalletStore.ts
```diff
@@ -30,6 +36,25 @@ export default defineStore("WalletStore", {
...
+						daoIDStore.show_opt_in = !appIsOptedIn;
+						if (appIsOptedIn) {
+							openSuccessNotificationWithIcon(
+								"You have already Opted-in DAO App"
+							);
+						}
+					})
+					.catch((error) =>
+						openErrorNotificationWithIcon(
+							"Unsuccessful while getting DAO App Opt-in Details",
```
rmeena840: ditto

src/store/DaoID.ts
```diff
@@ -24,5 +25,13 @@ export default defineStore("DaoStore", {
 			this.available = undefined;
 			this.locked = undefined;
 		},
+		removeDaoID() {
```
rmeena840: Best name would be `resetDaoStore`

rmeena840 (approved):
Pre-approving. Check the comments before merge.

PR #35, opened 2022-05-12
Migrated from Algosigner algod to local algodClient

Megha-Dev-19:


src/indexer/index.ts
```diff
@@ -33,85 +33,53 @@ export const searchForAssets = async (
...
+		let localStateMap = undefined;
+		// parse local state of DAO app for current user to get their details related to app for UI
+		const localState = await indexerClient
+			.lookupAccountAppLocalStates(address)
+			.do();
+		const parsedLocalState = JSON.parse(JSON.stringify(localState));
+
+		if (parsedLocalState && parsedLocalState[LOCAL_STATE]) {
+			const applicationInfo = parsedLocalState[LOCAL_STATE].find(
+				(i: any) => i[APPLICATION_ID] === applicationId
```
vuvoth:
```suggestion
				(appInfo: any) => appInfo[APPLICATION_ID] === applicationId
```

src/indexer/index.ts
```diff
@@ -122,7 +90,7 @@ export const searchApplicationAndAccount = async () => {
 		return;
 	}
 	const dao_id = daoIdStore.dao_id;
-	const application = await searchForApplication(dao_id).catch((error) => {
+	const application = await getApplicationGlobalState(dao_id).catch((error) => {
```
vuvoth, line 94: What is different between `try..catch` and `await...catch`.

src/indexer/index.ts
```diff
@@ -183,45 +163,71 @@ export const isAssetOpted = async (
...
-		if (optedApplicationInfo && optedApplicationInfo["app-local-state"]) {
-			return optedApplicationInfo["app-local-state"].id === application_id;
+		let isApplicationOpted = false;
+		const applicationInfo = await indexerClient
+			.lookupAccountAppLocalStates(address)
+			.do();
+		const parsedApplicationInfo = JSON.parse(JSON.stringify(applicationInfo));
+		if (parsedApplicationInfo && parsedApplicationInfo[LOCAL_STATE]) {
+			const optedApplicationInfo = applicationInfo[LOCAL_STATE].find(
+				(i: any) => (i[APPLICATION_ID] = applicationId)
```
vuvoth:
```suggestion
				(appInfo: any) => (appInfo[APPLICATION_ID] = applicationId)
```

src/indexer/index.ts
```diff
@@ -183,45 +163,71 @@ export const isAssetOpted = async (
...
+export const getGovASATokenAmount = async (
+	address: string,
+	govtID: number
+): Promise<number> => {
+	try {
+		const assetInfo = await indexerClient.lookupAccountAssets(address).do();
+		let amount = 0;
+		if (assetInfo[ASSETS]) {
+			const govtAssetInfo = assetInfo[ASSETS].find(
+				(i: any) => i[ASSET_ID] === govtID
```
vuvoth:
```suggestion
				(assetInfo: any) => assetInfo[ASSET_ID] === govtID
```

src/indexer/index.ts
```diff
@@ -166,14 +142,18 @@ export const isAssetOpted = async (
...
-			path: `/v2/accounts/${address}/assets/${asset_id}`,
-		});
-		if (optedAssetInfo && optedAssetInfo["asset-holding"]) {
-			return optedAssetInfo["asset-holding"]["asset-id"] === asset_id;
+		let isAssetOpted = false;
+		const assetInfo = await indexerClient.lookupAccountAssets(address).do();
+
+		if (assetInfo.assets) {
+			const optedAssetInfo = assetInfo.assets.find(
+				(i: any) => i[ASSET_ID] === asset_id
```
vuvoth:
```suggestion
				(asset: any) => asset[ASSET_ID] === asset_id
```

src/indexer/index.ts
```diff
@@ -166,14 +142,18 @@ export const isAssetOpted = async (
 	asset_id: number
```
vuvoth: I recommend use camel case style.

src/indexer/index.ts
```diff
@@ -183,45 +163,71 @@ export const isAssetOpted = async (
...
-			throw e;
 		}
+		return amount;
+	} catch (e) {
+		console.log(e);
+		throw e;
 	}
 };
+
+export async function decodeStateMap(
```
vuvoth: We don't use any `await` in decodeStateMap. Can we remove async keyword?

src/indexer/index.ts
```diff
@@ -183,45 +163,71 @@ export const isAssetOpted = async (
...
+	state: Array<any>
+): Promise<Map<Key, StateValue>> {
+	const stateMap = new Map<Key, StateValue>();
+	if (state) {
+		for (const g of state) {
+			const key = Buffer.from(g.key, "base64").toString();
+			if (g.value.type === 1) {
```
vuvoth: What is `1` meaning in this case?

vuvoth (commented):
Nice work! Left few comments.

src/indexer/index.ts
Megha-Dev-19: actually there are two ways to handle asynchronous promise either by using async, await/catch or by then/catch 
async, await was introduced later in ES7.
you can look at it https://www.smashingmagazine.com/2020/11/comparison-async-await-versus-then-catch/

src/indexer/index.ts
Megha-Dev-19: Sure, will update

src/indexer/index.ts
Megha-Dev-19: yes, will update

src/indexer/index.ts
Megha-Dev-19: it signifies that it is bytes type not uint

vuvoth (approved):

PR #37, opened 2022-05-21
GraphQL backend and frontend integration.

rmeena840:
GraphQL backend and frontend integration.


backend/server.ts
```diff
@@ -0,0 +1,62 @@
...
+const QueryRoot = new GraphQLObjectType({
+	name: "Query",
+	fields: () => ({
+		Daos: {
+			type: new GraphQLList(DaoType),
+			description: "Dao Records",
+			resolve: async () => {
+				const res = await executeQuery(QUERY_GET_ALL_DAOS);
+				if (res && res.rows && res.rows.length) {
+					res.rows.map((item: any) => {
```
Megha-Dev-19: We can create a type of item since we know the structure of data

src/api/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const BACKEND_URL =
+	process.env.BACKEND_URL || "http://localhost:4000/api";
```
Megha-Dev-19, line 2: api is little ambiguous here

src/api/fetch.ts
```diff
@@ -0,0 +1,21 @@
...
+	return fetch(BACKEND_URL, {
+		method: "POST",
+		headers: {
+			"Content-Type": "application/json",
+			Accept: "application/json",
+		},
+		body: JSON.stringify({
+			query,
+		}),
+	})
```
Megha-Dev-19, line 13: Can you please explain me what we are doing here?

Megha-Dev-19 (commented):
Good work. Left some comments

src/api/fetch.ts
rmeena840: We are trying to call the backend API here with the body being passed as a query param. Similar to what we do in axios.

src/api/constants.ts
rmeena840: Graphql has only one endpoint.

backend/psql/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const POSTGRES_UNAME = process.env.POSTGRES_UNAME || "algorand";
```
vuvoth: Should we rename it to `config.ts' ?

backend/server.ts
```diff
@@ -0,0 +1,62 @@
+import { executeQuery } from "./psql/queryExecute";
+import { QUERY_GET_ALL_DAOS } from "./psql/query";
+const express = require("express");
```
vuvoth, line 3: Shoud we change code to use `import..from` ?

backend/server.ts
```diff
@@ -0,0 +1,62 @@
...
+
+app.use(
+	"/api",
+	graphqlHTTP({
+		schema: schema,
+		graphiql: true,
+	})
+);
+
+app.listen(4000);
```
vuvoth: I think the port should be import from env.
```suggestion
app.listen(PORT, () => {
    console.log(`Backend is listening at ${PORT}`);
});
```

backend/server.ts
```diff
@@ -0,0 +1,62 @@
...
+		app_id: { type: GraphQLInt, description: "app id" },
+		app_params: {
+			type: GraphQLString,
+			description: "app params like global state etc.",
+		},
+		asset_id: { type: GraphQLInt, description: "asset id" },
+	}),
+});
+
+const QueryRoot = new GraphQLObjectType({
```
vuvoth: Move GraphQL define and config to new folder( how about move it to `graphql`).

vuvoth (commented):
Nice work! 
I left few suggestions. 
Please check it! Thanks!

backend/server.ts
rmeena840: `DoaItemType` has been added.

backend/server.ts
rmeena840: No, it needs require.

backend/psql/constants.ts
rmeena840: Done

backend/server.ts
rmeena840: Good suggestion!!

backend/server.ts
rmeena840: Done

backend/server.ts
```diff
@@ -0,0 +1,26 @@
+import { BACKEND_PORT } from "./config";
+import { schema } from "./graphql/schema";
+const express = require("express");
+const cors = require("cors");
+const app = express();
+const { graphqlHTTP } = require("express-graphql");
+const {
+	GraphQLObjectType,
+	GraphQLSchema,
```
vuvoth: Remove unused value here.

vuvoth (approved):
LGTM

backend/server.ts
rmeena840: Oops, thx done

src/api/constants.ts
Megha-Dev-19: okay, thanks for informing

PR #38, opened 2022-05-25
Dao page UI

Megha-Dev-19:
image

image

image


backend/psql/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const POSTGRES_UNAME = process.env.POSTGRES_UNAME || "algorand";
```
rmeena840: rebase your branch with master branch. this should not be here. It's been moved to config.ts

src/indexer/index.ts
```diff
@@ -230,3 +231,35 @@ export function decodeStateMap(state: Array<any>): Map<Key, StateValue> {
...
+		}
+	}
+	return stateMap;
+}
+
+export async function getAssetInformation(
+	assetId: number
+): Promise<UnknownObject> {
+	try {
+		const assetInfo = await indexerClient.lookupAssetByID(assetId).do();
```
rmeena840, line 259: Shouldn't we check if the value is present in it before? Array gives runtime error if the index value is not present.

src/pages/AllDao.vue
```diff
@@ -1,110 +1,282 @@
...
+				}"
+			>
+				<div class="padding_sm">
+					<a-input
+						ref="searchInput"
+						:placeholder="`Search ${column.dataIndex}`"
+						:value="selectedKeys[0]"
+						class="search_input"
+						@change="
+							(e) => setSelectedKeys(e.target.value ? [e.target.value] : [])
```
rmeena840: Move it to method block.

src/pages/AllDao.vue
```diff
@@ -1,110 +1,282 @@
...
+					title: "Dao App ID",
+					key: "dao_id",
+					dataIndex: "dao_id",
+				},
+				{
+					title: "Name",
+					key: "name",
+					dataIndex: "name",
+					customFilterDropdown: true,
+					onFilter: (value: string, record: DaoTableData) => {
```
rmeena840: We should move this logic to method block.

src/pages/AllDao.vue
```diff
@@ -1,110 +1,282 @@
...
 	},
 	setup() {
 		const formState = reactive(DaoStore());
+		const state = reactive({
+			searchText: "",
+			searchedColumn: "",
+		});
+		const searchInput = ref();
+		const tempArray: Array<DaoTableData> = [];
+		formState.psqlData = tempArray; // prohibit duplication of data
```
rmeena840, line 247: We can move this inside at line 243. First push all in temp array and then make psqlData equal to temp array. In this way, if API fails, we will retain the existing data.

src/pages/Proposal.vue
```diff
@@ -1,142 +1,137 @@
...
-			<a-col :span="12" :offset="6">
-				<div v-if="error" class="margin_bottom_sm">
-					<a-alert
-						message="Error"
-						:description="error"
-						type="error"
-						show-icon
-						closable
-						@close="error = ''"
+	<a-row>
```
rmeena840, line 2: Why we are changing here for All dao?

src/types/enum.types.ts
```diff
@@ -34,7 +34,7 @@ export enum VoteOptions {
 export enum EndPoint {
 	VOTE = "/vote",
 	ADD_PROPOSAL = "/addProposal",
-	ALL_DAO = "/allDao",
+	ALL_DAO = "/",
```
rmeena840, line 37: Why???

rmeena840 (changes requested):
Reviewed

src/components/AppHeader.vue
```diff
@@ -27,11 +20,19 @@
...
+								@click="() => handleMenuClick(NavigationKey.CREATE_DAO)"
+								>Create DAO</a-button
+							>
+
 							<router-link :to="{ path: EndPoint.ADD_PROPOSAL }">
 								<a-button
 									class="menu_option"
 									:type="isLinkActive(NavigationKey.ADD_PROPOSAL)"
 									@click="() => handleMenuClick(NavigationKey.ADD_PROPOSAL)"
+									:disabled="!daoID || showOptIn"
```
vuvoth: I see `!daoID || showOptIn` repeat few times. Should we create variable to store it in data?

vuvoth (approved):

src/indexer/index.ts
Megha-Dev-19: It will exist since it is there in the dao table that means it is deployed, also we are not checking for account specific asset data rather we are fetching for asset info itself

src/components/AppHeader.vue
Megha-Dev-19: yes, will update

src/pages/Proposal.vue
Megha-Dev-19: actually before we used to show an error on the screen if app was not opted in or dao id was not selected, but now instead of showing the error we disable the action(i.e for proposal, vote tokens, vote)

src/types/enum.types.ts
Megha-Dev-19: usually the first/starting page is assigned as / so I changed `allDao` it to /  as it will be the first page.
If /allDao makes more sense I can update it as well, what do you think ?

robert-zaremba:
We should still be able to see that the DAO ID cell in the table is editable when the DAO is set / selected.

Megha-Dev-19:
> We should still be able to see that the DAO ID cell in the table is editable when the DAO is set / selected.

@robert-zaremba  Does this work? 

image
image

src/pages/AllDao.vue
Megha-Dev-19: Will update it in pagination, since there we will need to clean the data with page change

robert-zaremba:
Yes, looks 👍

src/pages/AllDao.vue
rmeena840: Ok

src/pages/Proposal.vue
rmeena840: Ok

src/types/enum.types.ts
rmeena840: Ok

backend/psql/constants.ts
rmeena840: This file is still here???

src/indexer/index.ts
rmeena840: Ok

rmeena840 (approved):
Pre-approving it. Just make sure to get rid of that constant.ts file. It should not be here.

backend/psql/constants.ts
Megha-Dev-19: removed

PR #39, opened 2022-05-26
Backend support for pagination

rmeena840:
Backend support for pagination. Getting only the needed data instead of fetching all data from SQL DB.
Task:
https://www.pivotaltracker.com/story/show/182295713

### DIFF
--- a/backend/graphql/handler.ts
+++ b/backend/graphql/handler.ts
@@ -1,10 +1,10 @@
 import { executeQuery } from "../psql/queryExecute";
 import { QUERY_GET_DAOS_COUNT, getPaginatedQuery } from "../psql/query";
-import { DaosAndPageInfoType } from "./types";
-import { DaoItemType, DaoArgType, DaoAndPageResType } from "../types";
+import { DaosAndPageType } from "./types";
+import { DaoItemType, Page, DaosAndPageResType } from "../types";
 const { GraphQLObjectType, GraphQLInt, GraphQLNonNull } = require("graphql");
 
-const validateRes = (totalDaosRes: any, paginateRes: any, args: DaoArgType) => {
+const validateRes = (totalDaosRes: any, paginateRes: any, args: Page) => {
 	return (
 		totalDaosRes &&
 		totalDaosRes.rows &&
@@ -18,15 +18,15 @@ const validateRes = (totalDaosRes: any, paginateRes: any, args: DaoArgType) => {
 const getDaoPageRes = (
 	totalDaos: number,
 	daos: [DaoItemType],
-	args: DaoArgType
-): DaoAndPageResType => {
+	args: Page
+): DaosAndPageResType => {
 	const pageNumber = args.pageNumber; // page number = 1..n
 	const pageSize = args.pageSize;
 	const hasNext = !(pageNumber * pageSize >= totalDaos); // check if user's total page iterated is >= totalDaos
 	const hasPrev = !(pageNumber <= 1); // check if user is on first page
 	return {
 		daos,
-		pageInfo: {
+		PageInfo: {
 			hasNext,
 			hasPrev,
 			totalDaos,
@@ -38,7 +38,7 @@ export const queryRoot = new GraphQLObjectType({
 	name: "Query",
 	fields: () => ({
 		DaoAndPage: {
-			type: DaosAndPageInfoType,
+			type: DaosAndPageType,
 			description: "Returns Dao Records and Page Info",
 			args: {
 				pageNumber: {
@@ -50,7 +50,7 @@ export const queryRoot = new GraphQLObjectType({
 					description: "Offset as an argument.",
 				},
 			},
-			resolve: async (_: object, args: DaoArgType) => {
+			resolve: async (_: object, args: Page) => {
 				const totalDaosRes = await executeQuery(QUERY_GET_DAOS_COUNT);
 				const paginateRes = await executeQuery(
 					getPaginatedQuery(args.pageNumber, args.pageSize)
--- a/backend/graphql/handler.ts
+++ b/backend/graphql/handler.ts
@@ -1,25 +1,68 @@
 import { executeQuery } from "../psql/queryExecute";
-import { QUERY_GET_ALL_DAOS } from "../psql/query";
-import { DaoType } from "./types";
-import { DaoItemType } from "../types";
-const { GraphQLObjectType, GraphQLList } = require("graphql");
+import { QUERY_GET_DAOS_COUNT, QUERY_PAGINATED_DAOS } from "../psql/query";
+import { DaosAndPageInfoType } from "./types";
+import { DaoItemType, DaoArgType, DaoAndPageResType } from "../types";
+const { GraphQLObjectType, GraphQLInt, GraphQLNonNull } = require("graphql");
+
+const validateRes = (totalDaosRes: any, paginateRes: any, args: DaoArgType) => {
+	return (
+		totalDaosRes &&
+		totalDaosRes.rows &&
+		paginateRes &&
+		paginateRes.rows &&
+		paginateRes.rows.length &&
+		args
+	);
+};
+
+const getDaoPageRes = (
+	totalDaos: number,
+	Daos: [DaoItemType],
+	args: DaoArgType
+): DaoAndPageResType => {
+	const pageNumber = args.pageNumber; // page number = 1..n
+	const pageSize = args.pageSize;
+	const hasNext = !(pageNumber * pageSize >= totalDaos); // check if user's total page iterated is >= totalDaos
+	const hasPrev = !(pageNumber <= 1); // check if user is on first page
+	return {
+		Daos,
+		pageInfo: {
+			hasNext,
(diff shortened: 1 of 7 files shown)


backend/graphql/handler.ts
```diff
@@ -1,25 +1,68 @@
...
-import { QUERY_GET_ALL_DAOS } from "../psql/query";
-import { DaoType } from "./types";
-import { DaoItemType } from "../types";
-const { GraphQLObjectType, GraphQLList } = require("graphql");
+import { QUERY_GET_DAOS_COUNT, QUERY_PAGINATED_DAOS } from "../psql/query";
+import { DaosAndPageInfoType } from "./types";
+import { DaoItemType, DaoArgType, DaoAndPageResType } from "../types";
+const { GraphQLObjectType, GraphQLInt, GraphQLNonNull } = require("graphql");
+
+const validateRes = (totalDaosRes: any, paginateRes: any, args: DaoArgType) => {
```
vuvoth: Could we use type on those arguments?

src/pages/AllDao.vue
```diff
@@ -86,16 +86,18 @@ export default defineComponent({
 	setup() {
 		const formState = reactive(DaoStore());
 		// Get all daos
-		executeReq(ALL_DAOS_REQ)
+		// Here 1 --> Page number
+		// 5 --> number of entries to show in single fetch
+		executeReq(ALL_DAOS_REQ(1, 5))
```
vuvoth: I think we should save `1, 5` as constants.

backend/graphql/handler.ts
```diff
@@ -1,25 +1,68 @@
...
+				},
+				pageSize: {
+					type: new GraphQLNonNull(GraphQLInt),
+					description: "Offset as an argument.",
+				},
+			},
+			resolve: async (_: object, args: DaoArgType) => {
+				const totalDaosRes = await executeQuery(QUERY_GET_DAOS_COUNT);
+				const paginateRes = await executeQuery(
+					QUERY_PAGINATED_DAOS(args.pageNumber, args.pageSize)
```
robert-zaremba: why uppercase? Uppercase should be only for constants

backend/graphql/handler.ts
```diff
@@ -1,25 +1,68 @@
...
+					type: new GraphQLNonNull(GraphQLInt),
+					description: "Page number as an argument",
+				},
+				pageSize: {
+					type: new GraphQLNonNull(GraphQLInt),
+					description: "Offset as an argument.",
+				},
+			},
+			resolve: async (_: object, args: DaoArgType) => {
+				const totalDaosRes = await executeQuery(QUERY_GET_DAOS_COUNT);
```
robert-zaremba, line 54: why do you query and return all DAOs?

robert-zaremba (commented):
I don't think this is correctly done.
1. we should avoid querying daos 2 times (currenty this is done in `queryRoot`)
2. Queries are not organized correctly, as they should be in a GraphQL project. See this: https://www.youtube.com/watch?v=kG5RQHhYEto

Can you check if we really need custom resolvers. I think GraphQL can work directly with postgres and understand postgresql scheme and pagination.

backend/graphql/handler.ts
rmeena840: Because we need information related to pagination. We need to check if prev or next exists or not.

backend/graphql/handler.ts
rmeena840: It will be `any`. We don't know the type.

src/pages/AllDao.vue
rmeena840: More work has to done here. I will add TODO.

backend/graphql/handler.ts
```diff
@@ -1,25 +1,68 @@
...
+		paginateRes &&
+		paginateRes.rows &&
+		paginateRes.rows.length &&
+		args
+	);
+};
+
+const getDaoPageRes = (
+	totalDaos: number,
+	Daos: [DaoItemType],
```
sczembor: Why uppercase?

backend/graphql/handler.ts
rmeena840: Done

backend/graphql/handler.ts
```diff
@@ -1,25 +1,69 @@
...
+
+const getDaoPageRes = (
+	totalDaos: number,
+	daos: [DaoItemType],
+	args: DaoArgType
+): DaoAndPageResType => {
+	const pageNumber = args.pageNumber; // page number = 1..n
+	const pageSize = args.pageSize;
+	const hasNext = !(pageNumber * pageSize >= totalDaos); // check if user's total page iterated is >= totalDaos
+	const hasPrev = !(pageNumber <= 1); // check if user is on first page
```
Megha-Dev-19, line 26: can we assign types here.?

rmeena840: By default, it will take a boolean.

backend/graphql/types.ts
```diff
@@ -12,3 +19,30 @@ export const DaoType = new GraphQLObjectType({
 		asset_id: { type: GraphQLInt, description: "asset id" },
 	}),
 });
+
+const PageInfoType = new GraphQLObjectType({
```
robert-zaremba:
```suggestion
const PageType = new GraphQLObjectType({
```

backend/types/interface.types.ts
```diff
@@ -3,3 +3,19 @@ export interface DaoItemType {
 	app_params?: string;
 	asset_id?: number;
 }
+
+export interface DaoArgType {
```
robert-zaremba:
```suggestion
export interface Page {
```

backend/types/interface.types.ts
```diff
@@ -3,3 +3,19 @@ export interface DaoItemType {
...
+	pageNumber: number;
+	pageSize: number;
+}
+
+interface pageInfo {
```
robert-zaremba:
```suggestion
interface PageInfo {
```

backend/psql/query.ts
```diff
@@ -1 +1,9 @@
-export const QUERY_GET_ALL_DAOS = "SELECT * FROM sigma_daos";
+export const QUERY_GET_DAOS_COUNT = "SELECT COUNT(*) FROM sigma_daos";
+
+/**
+ * Paginated query to fetch only the required information from database instead of all
+ * @param pageNumber Current page number to fetch data for.
+ * @param pageSize Number of data to show in single page/fetch.
+ */
+export const getPaginatedQuery = (pageNumber: number, pageSize: number) =>
```
robert-zaremba: the name is too general, we are DAOs. Let's rename this function.

backend/graphql/handler.ts
```diff
@@ -1,25 +1,69 @@
...
+			resolve: async (_: object, args: DaoArgType) => {
+				const totalDaosRes = await executeQuery(QUERY_GET_DAOS_COUNT);
+				const paginateRes = await executeQuery(
+					getPaginatedQuery(args.pageNumber, args.pageSize)
+				);
+				if (validateRes(totalDaosRes, paginateRes, args)) {
+					const totalDaos = totalDaosRes.rows[0].count || 0;
+					paginateRes.rows.map((item: DaoItemType) => {
 						if (item.app_params) {
 							item.app_params = JSON.stringify(item.app_params);
```
robert-zaremba, line 63: why do we call `stringify` here?

backend/types/interface.types.ts
```diff
@@ -3,3 +3,19 @@ export interface DaoItemType {
...
+	hasPrev: boolean;
+	hasNext: boolean;
+	totalDaos: number;
+}
+
+export interface DaoAndPageResType {
```
robert-zaremba:
```suggestion
export interface DaosAndPageResType {
```

robert-zaremba (commented):
please look again at names and functions to make sure they are clear enough and meaningful enough.
+ I think we can remove `getDaoPageRes` or rename it to `combineDaosAndPage`

backend/graphql/handler.ts
rmeena840: app_param field in DB has JSONB as type. GraphQL was not having this type so I converted this into string format and returned it. This way it will match the return type.
Check this: https://github.com/scale-it/sigmadao-ui/blob/4a81c8b8f889157283d4d481055e9387f25cdbce/backend/graphql/types.ts#L15
It returns return type in string format.

backend/psql/query.ts
rmeena840: ok. converted to `getPageQuery`

robert-zaremba: you should rename to something with DAO in name, `getPaginatedDaos` for example.

backend/graphql/handler.ts
robert-zaremba: then you should convert it to an object, we don't want to send stringified objects.

rmeena840: This will be tackled in postgraphile task that you have created.

robert-zaremba (approved):

PR #42, opened 2022-06-02
Package upgrade

Megha-Dev-19:
Upgrade algo-builder version
Now we can use privateNet with all three wallets(i.e AlgoSigner, MyAlgo Wallet and Wallet Connect)

### DIFF
--- a/src/components/WalletConnect.vue
+++ b/src/components/WalletConnect.vue
@@ -98,7 +98,7 @@
 <script lang="ts">
 import { DownOutlined, LogoutOutlined } from "@ant-design/icons-vue";
 import { defineComponent } from "vue";
-import { WalletType, NetworkTypes } from "@/types";
+import { WalletType, NetworkTypes, HttpNetworkConfig } from "@/types";
 import {
 	MyAlgoWalletSession,
 	WallectConnectSession,
@@ -106,12 +106,18 @@ import {
 } from "@algo-builder/web";
 import WalletStore from "../store/WalletStore";
 import { searchApplicationAndAccount } from "@/indexer";
-import { walletMessage } from "../constants/constant";
+import {
+	mainnetURL,
+	betanetURL,
+	testnetURL,
+	walletMessage,
+} from "../constants/constant";
 import DaoID from "@/store/DaoID";
 import {
 	openErrorNotificationWithIcon,
 	WALLET_CONNECTION_ERROR,
 } from "@/constants";
+import { port, server, token } from "@/config/algob.config";
 declare var AlgoSigner: any; // eslint-disable-line
 
 export default defineComponent({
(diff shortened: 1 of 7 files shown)


src/constants/constant.ts
```diff
@@ -65,3 +65,7 @@ export const walletMessage = {
 	NETWORK_ISSUE: (walletType: string) =>
 		`Please select a network to connect with ${walletType}`,
 };
+
+export const mainnetURL = "https://node.algoexplorerapi.io";
```
vuvoth: constant should have uppercase format.

src/types/interface.types.ts
```diff
@@ -60,3 +60,18 @@ export interface DaoTableData {
 export interface UnknownObject {
 	[key: string]: string | number | boolean;
 }
+
+export interface AlgodTokenHeader {
```
vuvoth: Could you import those types from algo-builder?

vuvoth (commented):
quick review!

src/constants/constant.ts
Megha-Dev-19: oh yes will update

src/types/interface.types.ts
Megha-Dev-19: I will check if we export it from there, will update

src/components/WalletConnect.vue
```diff
@@ -106,12 +106,19 @@ import {
...
+	MAIN_NET_URL,
+	BETA_NET_URL,
+	TEST_NET_URL,
+	walletMessage,
+} from "../constants/constant";
 import DaoID from "@/store/DaoID";
 import {
 	openErrorNotificationWithIcon,
 	WALLET_CONNECTION_ERROR,
 } from "@/constants";
```
vuvoth, line 117: What is different between `../constants/constant` and `@constants` ?

vuvoth, line 119: What is different between `../constants/constant` and `@constants` ?

vuvoth (approved):
Pre-approved! Thanks for your PR ~~

src/components/WalletConnect.vue
rmeena840: You have tried to use another import from same folder. `"../constants/constant"` and `@/constants`. Please use import. I would recommend 2nd one.

rmeena840 (changes requested):
Reviewed.

src/components/WalletConnect.vue
Megha-Dev-19: `@constants` tells the folder from which to import i.e it's `index.tx` while `../constants/constant` is the specific address of the file

Megha-Dev-19: yeah you are right, I missed it, will update

rmeena840 (approved):
lgtm

PR #44, opened 2022-06-09
Migrate from indexer client and use postgraphile for fetching data from PostgreSQL

rmeena840:
Migrate from indexer client and use postgraphile for fetching data from PostgreSQL
Task: https://www.pivotaltracker.com/story/show/182318157


src/pages/AllDao.vue
```diff
@@ -362,7 +362,7 @@ export default defineComponent({
 							key: index,
 							dao_id: +item.appId,
 							token_id: +item.assetId,
-							token_name: tokenData.name as string,
+							token_name: tokenData.an as string,
```
Megha-Dev-19, line 365: Why is it `an`

src/pages/Proposal.vue
```diff
@@ -245,9 +245,9 @@ export default defineComponent({
 						}
 					}
 
-					// check if asset is already opted
+					// check if app is already opted
 					const isApplicationAlreadyOpted = await isApplicationOpted(
-						this.walletStore.address,
+						lsig.address(),
```
Megha-Dev-19, line 250: Why are we using lsig here, shouldn't it be specific to user connected, to check if he/she has opted the app

src/utility/algod.utility.ts
```diff
@@ -91,3 +91,8 @@ export const optInDaoApp = async (
 		throw error;
 	}
 };
+
+export const convertToHex = (address: string) => {
+	const pk = algosdk.decodeAddress(address);
+	return Buffer.from(pk.publicKey).toString("hex");
+};
```
Megha-Dev-19, line 98: I didn't understand the use of this function, what does it do.?

src/pages/AllDao.vue
rmeena840: We are now directly using the PostreSQL for fetching data. Indexer Client used to do these proper naming before it returns the response on client side but now we are fetching directly. This is how it is stored in PostreSQL.

src/utility/algod.utility.ts
rmeena840: It converts the address into a hexadecimal address. The address is stored in PostgreSQL in hexadecimal encoding.

src/pages/Proposal.vue
rmeena840: I have checked the add propsal script. It checks if it has been opted by lsig or not. So lsig will be here not the wallet address.

`// check if app is already opted
					const isApplicationAlreadyOpted = await isApplicationOpted(
						lsig.address(),
						this.daoStore.dao_id as number
					);
					// optin
					if (!isApplicationAlreadyOpted) {
						await this.optInLsigToApp(lsig);
					}`

check here we are trying to opt in lsig

src/pages/AllDao.vue
Megha-Dev-19: okay, thanks for the information

src/utility/algod.utility.ts
Megha-Dev-19: okay, I see that PostgreSQL stores data quite differently than indexer

src/pages/Proposal.vue
Megha-Dev-19: okay

Megha-Dev-19 (approved):
Lgtm

src/pages/AllDao.vue
robert-zaremba, line 365: let's try to avoid using `as`

PR #45, opened 2022-06-13
Updated sequence of wallet and network input

Megha-Dev-19:
Now user needs to provide wallet type first and then select the network type.

### DIFF
--- a/src/components/WalletConnect.vue
+++ b/src/components/WalletConnect.vue
@@ -17,13 +17,17 @@
 						</a-menu>
 					</template>
 					<a-button>
-						Connect Wallet
+						<span v-if="selectedWallet === WalletType.NONE">
+							Connect Wallet</span
+						>
+						<span v-else>Selected Wallet : {{ selectedWallet }}</span>
+
 						<DownOutlined />
 					</a-button>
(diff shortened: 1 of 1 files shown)


src/components/WalletConnect.vue
```diff
@@ -4,43 +4,43 @@
...
-						>
-						<span v-else>Selected Network : {{ selectedNetwork }}</span>
+						Connect Wallet
 						<DownOutlined />
 					</a-button>
 				</a-dropdown>
 			</a-col>
 
-			<a-col v-if="selectedNetwork !== NetworkTypes.NONE">
+			<a-col v-if="selectedWallet !== 0">
```
vuvoth: Could we use something like walletType.NONE instead of "0" at here?

vuvoth (approved):
I tried it on my laptop. Worked very well! 
I approved. You should wait another approve before merge PR. 
Nice work @Megha-Dev-19!

rmeena840 (approved):

src/components/WalletConnect.vue
Megha-Dev-19: oh yes, updated

PR #51, opened 2022-06-21
Update: Sigma Dao README

rmeena840:
Update: Sigma Dao README
Task: https://www.pivotaltracker.com/story/show/182424903

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,24 @@
-# Nuxt 3 Minimal Starter
+<img src="./src/assets/logo.png" height="200" title="Sigma Dao" />
 
-We recommend to look at the [documentation](https://v3.nuxtjs.org).
+# Sigma Dao
+
+Sigma Dao is a web application that let's an user create a dao, add proposal, deposit tokens and vote.
+
+Wallets supported by Sigma Dao are:
+
+- [AlgoSigner](https://www.purestake.com/technology/algosigner/)
+- [MyAlgo Wallet](https://wallet.myalgo.com/new-account)
+- [Wallet Connect](https://walletconnect.com/)
+
+## Requirements
+
+- Node version >=14.11.0
+- yarn 3.2
+- PostgreSQL >= 12
 
 ## Setup
 
-Make sure to install the dependencies
+Install the dependencies
 
 ```bash
 yarn install
@@ -18,6 +32,12 @@ Start the development server on http://localhost:3000
 yarn dev
 ```
 
+Start the backend development server on http://localhost:4000/graphiql
+
+```bash
+yarn server
+```
+
 ## Production
 
 Build the application for production:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,6 +32,12 @@ Start the development server on http://localhost:3000
 yarn dev
 ```
 
+Start the backend development server on http://localhost:4000/graphiql
```
vuvoth, line 41: How about setup indexer for Sigma Dao?

README.md
```diff
@@ -1,10 +1,24 @@
-# Nuxt 3 Minimal Starter
+<img src="./src/assets/logo.png" height="200" title="Sigma Dao" />
 
-We recommend to look at the [documentation](https://v3.nuxtjs.org).
+# Sigma Dao
+
```
vuvoth, line 4: Should we add "beta software" warning, @robert-zaremba ?

README.md
rmeena840: We are not setting up indexer from Sigma Dao. We have removed the indexer client from sigma dao.

vuvoth: No, I mean for deploy DAO approval app and setup db for graph ql.

rmeena840: Yes, correct. We should add it. I have added it now. I would prefer link the `examples/dao` README. This would result in less maintenance of this README in case if examples/dao readme updates with steps, which is very much likely to happen.

README.md
```diff
@@ -18,6 +36,12 @@ Start the development server on http://localhost:3000
 yarn dev
 ```
 
+Start the backend development server on http://localhost:4000/graphiql
+
```
Megha-Dev-19, line 42: may be we can add a little bit more information about the server, like it connects with the psql database which you set up using examples/dao/readme

README.md
```diff
@@ -1,15 +1,33 @@
-# Nuxt 3 Minimal Starter
+<img src="./src/assets/logo.png" height="200" title="Sigma Dao" />
 
-We recommend to look at the [documentation](https://v3.nuxtjs.org).
+# Sigma Dao
+
+Sigma Dao is a web application that let's an user create a dao, add proposal, deposit tokens and vote.
```
Megha-Dev-19:
```suggestion

Sigma Dao is a web application that let's a user create a DAO, enable them to Add Proposal, Deposit and Withdraw Tokens and also allow them to Vote.
```

README.md
rmeena840: Done

README.md
robert-zaremba: Yes

rmeena840: I have added a beta version note at the end.

vuvoth (approved):
Nice work! Thanks @rmeena840!

PR #52, opened 2022-06-22
Update dao appID query on frontend

Megha-Dev-19:


src/indexer/index.ts
```diff
@@ -262,3 +268,43 @@ export async function getAssetInformation(
 		throw e;
 	}
 }
+
+/**
+ * Get parsed DAO information
+ * @param params object to decode
+ */
+export async function decodePSQLAppParams(params: any): Promise<DaoTableData> {
```
rmeena840:
```suggestion
export async function decodeAppParams(params: any): Promise<DaoTableData> {
```
PSQL is not related to app params. Some other name is needed.

src/indexer/index.ts
```diff
@@ -262,3 +268,43 @@ export async function getAssetInformation(
 		throw e;
 	}
 }
+
+/**
+ * Get parsed DAO information
+ * @param params object to decode
```
rmeena840:
```suggestion
 * @param params object to be decoded
```

src/indexer/index.ts
```diff
@@ -262,3 +268,43 @@ export async function getAssetInformation(
...
+		link: globalState.get(GLOBAL_STATE_MAP_KEY.Url) as string,
+	};
+}
+
+/**
+ * Get dao information
+ * @param searchDaoType parameter to query
+ * @param value paramter value
+ */
+export async function handleDaoSearch(
```
rmeena840, line 294: I would suggest to keep this method in AllDao.vue file since it will be used there only. Other component won't be using it.

rmeena840 (changes requested):
Reviewed

src/indexer/index.ts
Megha-Dev-19: actually it will be used in AllDao as well as AppHeader, currently I have used it in AppHeader only

src/indexer/index.ts
Megha-Dev-19: I have a function for `decodeAppParamsState` so I added PSQL to make it separate that the whole object received is decoded

rmeena840 (approved):
Pre-approving. Look the comments and other that we discussed.

PR #53, opened 2022-06-22
Added 32 characters check to dao name

Megha-Dev-19:


src/pages/CreateDao.vue
```diff
@@ -98,7 +98,7 @@
 					name="dao_name"
 					:rules="[{ required: true }]"
 				>
-					<a-input v-model:value="formState.dao_name" />
+					<a-input maxlength="32" v-model:value="formState.dao_name" />
```
rmeena840, line 101: Quick question.
What will happen here if dao_name is being returned greater then 32 char size? Will it be blank?

rmeena840 (approved):

src/pages/CreateDao.vue
Megha-Dev-19: it won't allow user to input any letter exceeding that

rmeena840: ok

PR #54, opened 2022-06-23
updated contract

Megha-Dev-19:
- Updated contract teal code of govt_asset_id check

### DIFF
--- a/src/contract/dao/dao-app-approval.ts
+++ b/src/contract/dao/dao-app-approval.ts
@@ -1,1121 +1,1120 @@
 import { compileToUnit8Array } from "@/utility";
 
 const daoAppApproval = `#pragma version 6
-txn ApplicationID
-int 0
-==
-bnz main_l86
-txn OnCompletion
-int UpdateApplication
-==
-txn OnCompletion
-int DeleteApplication
-==
-||
-bnz main_l85
-txn OnCompletion
-int CloseOut
-==
-txn OnCompletion
-int OptIn
-==
-||
-bnz main_l84
-txna ApplicationArgs 0
-byte "optin_gov_token"
-==
-bnz main_l83
-txna ApplicationArgs 0
-byte "add_proposal"
-==
-bnz main_l72
-txna ApplicationArgs 0
-byte "deposit_vote_token"
-==
-bnz main_l71
-txna ApplicationArgs 0
-byte "register_vote"
-==
-bnz main_l56
-txna ApplicationArgs 0
-byte "execute"
-==
-bnz main_l39
-txna ApplicationArgs 0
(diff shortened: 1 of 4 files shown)


src/pages/CreateDao.vue
```diff
@@ -222,6 +222,7 @@ export default defineComponent({
 						approvalProgramBytes: await getCompiledDaoApproval(),
 						clearProgramBytes: await getCompiledDaoClear(),
 						...DAO_CONTRACT_STATE_CONFIG,
+						foreignAssets: [token_id as number],
```
ratik21, line 225: why do we need `as number` here?

src/contract/dao/dao-app-approval.ts
```diff
@@ -1,1115 +1,1120 @@
...
+app_global_put
+byte "dao_name"
+txna ApplicationArgs 5
+app_global_put
+byte "gov_token_id"
+txna ApplicationArgs 6
+btoi
+app_global_put
+int 1
+return
```
ratik21: I think in general we should keep the separate DAO contract file (pyTEAL file) here, and use the `PyCompileOp` here to compile that contract to TEAL.

ratik21 (approved):

src/pages/CreateDao.vue
Megha-Dev-19: when I assign the type, it can be undefined also (when we initialize the state), but here it is definite that it will exist so to get away with the warning used `as number`

src/contract/dao/dao-app-approval.ts
Megha-Dev-19: Will create a separate task for it, as we have multiple files which should be updated

vuvoth (approved):

src/contract/dao/dao-app-approval.ts
```diff
@@ -1,1115 +1,1120 @@
 import { compileToUnit8Array } from "@/utility";
 
 const daoAppApproval = `#pragma version 6
```
rmeena840: @Megha-Dev-19 The previous single tab indentation was correct. Please add single tab indentation.

rmeena840 (approved):
Pre-approving. Please look at the comments.

rmeena840:
@Megha-Dev-19 Have you also checked the other contract files present in examples/dao of algobuilder? We should all contract files in sigma dao with latest contract update on examples/dao of algo-builder.

Megha-Dev-19:
> @Megha-Dev-19 Have you also checked the other contract files present in examples/dao of algobuilder? We should all contract files in sigma dao with latest contract update on examples/dao of algo-builder.

Yeah, I checked other files are updated. Thanks for reminding

PR #56, opened 2022-06-27
Add: Add search by dao name backend support for Sigma Dao with pagination

rmeena840:
Add search by dao name backend support for Sigma Dao with pagination
Task: https://www.pivotaltracker.com/story/show/182496847

### DIFF
--- a/src/api/query.request.ts
+++ b/src/api/query.request.ts
@@ -134,3 +134,43 @@ export const getDaoInfoByAppIdReq = (appId: number) => {
     }
   }`;
 };
+
+/** SearchByDaoName
+ * @param daoName Dao name to be searched
+ * @param first Similar to limit in PostreSQL. E.g first 10 entries of row after cursor pointer
+ * @param after End cursor pointer
+ * @param last  Similar to limit in PostreSQL. E.g last 10 entries of row before cursor pointer
+ * @param before Start cursor pointer
+ */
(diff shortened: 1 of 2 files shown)


src/api/query.request.ts
```diff
@@ -134,3 +134,43 @@ export const getDaoInfoByAppIdReq = (appId: number) => {
...
+ * @param after End cursor pointer
+ * @param last  Similar to limit in PostreSQL. E.g last 10 entries of row before cursor pointer
+ * @param before Start cursor pointer
+ */
+export const getDaoInfoByAppNameReq = (
+	daoName: string,
+	first: number | null,
+	after: string | null,
+	last: number | null,
+	before: string | null
```
Megha-Dev-19, line 150: so even with the backend we will be querying only for current page data not complete data. just confirming.? because I think if user is searching for dao name he/she would expect the entire list.

rmeena840: Yes, search result can be very large. This also has to be paginated on frontend side.
E.g There are 100 daos with name 'DAO' then if user searches for just 'd' then backend will throw 100 results. Hence, need to be paginated.

Megha-Dev-19: okay

Megha-Dev-19 (approved):
Lgtm

PR #57, opened 2022-06-29
Add: Create a Proposal List in SigmaDAO

rmeena840:
Screenshot 2022-07-06 at 12 13 00 PM


src/components/ProposalTable.vue
```diff
@@ -0,0 +1,246 @@
...
+	<div class="dao_table_container">
+		<a-table
+			:dataSource="dataSource"
+			:columns="columns"
+			bordered
+			:pagination="false"
+		>
+			<template #title>
+				<a-col>
+					<h3 style="text-align: center">Available Proposals</h3>
```
vuvoth, line 27: I worry about Completed Proposal and close Proposal. Could we see it?

rmeena840: This functionality has not been yet and is also not in the scope of this PR and task. I will create a task for this.

src/components/ProposalTable.vue
```diff
@@ -0,0 +1,246 @@
...
+				getProposalCursorReq(appId, pageNumber, ROWS_PER_PAGE)
+			).catch((error) =>
+				openErrorNotificationWithIcon(UNSUCCESSFUL, error.message)
+			);
+
+			if (
+				cursorRes &&
+				cursorRes.searchSigmaDaosProposals &&
+				cursorRes.searchSigmaDaosProposals.pageInfo
+			) {
```
Megha-Dev-19: Can we use nullish coalescing here?

src/constants/constant.ts
```diff
@@ -25,6 +25,20 @@ export const ProposalType = {
 	ALGO_TRANSFER: 1,
 	ASA_TRANSFER: 2,
 	MESSAGE: 3,
+	[1]: "Algo Transer",
```
Megha-Dev-19:
```suggestion
	[1]: "Algo Transfer",
```

src/indexer/index.ts
```diff
@@ -254,6 +259,30 @@ export function decodeAppParamsState(state: any): Map<Key, StateValue> {
 	return stateMap;
 }
 
+/**
+ * Decode local app params state
+ * @param state State to be decoded
+ */
+export function decodeLocalAppParamsState(state: any): Map<Key, StateValue> {
```
Megha-Dev-19: Don't we have a function to decode local and global state, can't that be used here?
If not can we modify existing one to use it dynamically

Megha-Dev-19 (changes requested):
Left few suggestions, which tab(Add proposal or vote) are we going to show this list and what do we plan on adding to its action..?

src/components/ProposalTable.vue
rmeena840: Yes, sure

src/indexer/index.ts
rmeena840: Thanks, I have used predefined method now.

rmeena840:
> Left few suggestions, which tab(Add proposal or vote) are we going to show this list and what do we plan on adding to its action..?

@robert-zaremba I would suggest to show proposal list in vote tab. Let me know your thought on this and also what are all the fields have to display. I have shared the current UI snaps in #dapp-ui channel. Please have a look and give us the UI preference.

robert-zaremba:
I agree @rmeena840

robert-zaremba:
Left UI suggestion in Discord. Thanks!

src/components/ProposalTable.vue
```diff
@@ -0,0 +1,242 @@
...
+export default defineComponent({
+	name: "ProposalTable",
+	data() {
+		return {
+			key: "ProposalKey",
+			columns: [
+				{
+					title: "Proposal Name",
+					key: "name",
+					dataIndex: "name",
```
Megha-Dev-19: I think we have `PROPOSAL_LOCAL_STATE_MAP_KEY` with these values, we should update to use the constants then

Megha-Dev-19 (changes requested):

src/components/ProposalTable.vue
rmeena840: Done.

src/components/AppHeader.vue
```diff
@@ -31,7 +31,7 @@
 								<a-button
 									class="menu_option"
 									:type="isLinkActive(EndPoint.ADD_PROPOSAL)"
-									:disabled="DaoStore().disableActions"
+									:disabled="!DaoStore().isDaoSelected"
```
Megha-Dev-19, line 34: why are we changing only this, we have `disableActions` other places as well

src/components/AppHeader.vue
```diff
@@ -139,7 +139,7 @@ export default defineComponent({
 		const walletStore = WalletStore();
 		return {
 			EndPoint,
-			daoID: daoStore.dao_id,
+			daoID: daoStore.searchDaoId,
```
Megha-Dev-19, line 142: what does searchDaoId means, is it the selected dao id by user?

src/components/AppHeader.vue
```diff
@@ -171,9 +171,10 @@ export default defineComponent({
 					null,
 					null
 				);
-
 				if (response) {
 					this.govtId = response.token_id;
+					this.name = response.name;
+					this.daoID = value;
```
Megha-Dev-19, line 177: I think we already we set it in `searchApplicationAndAccount`

src/components/ProposalTable.vue
```diff
@@ -0,0 +1,308 @@
...
+				<template v-if="column.key === 'name'">
+					<a :href="'//' + record.url" target="_blank">
+						{{ record.name }}
+					</a>
+				</template>
+				<template v-if="column.key === 'voting_start'">
+					{{
+						secToFormat(record.voting_start, "lll") +
+						" - " +
+						secToFormat(record.voting_end, "lll")
```
Megha-Dev-19: can we change it to `formatSecondsWithType`

src/components/ProposalTable.vue
```diff
@@ -0,0 +1,308 @@
...
+					<a-button type="link" @click="() => handleSelectProposal(record)">
+						Use
+					</a-button>
+					<a-button
+						type="link"
+						danger
+						@click="() => handleCloseProposal(record)"
+					>
+						Delete
+					</a-button>
```
Megha-Dev-19: should we display Close or Delete.? will that proposal will be deleted or closed for voting

src/components/ProposalTable.vue
```diff
@@ -0,0 +1,308 @@
...
+						handlePagination(
+							PaginationCallType.JUMP_PAGE,
+							originalElement.children[0].children
+						)
+					"
+					:is="originalElement"
+					v-else
+				></component>
+			</template>
+		</a-pagination>
```
Megha-Dev-19: I think we can have a reusable component for pagination, since it is used in multiple places now

src/components/ProposalTable.vue
```diff
@@ -0,0 +1,308 @@
...
+					first,
+					endCursor,
+					last,
+					startCursor
+				)
+			).catch((error) =>
+				openErrorNotificationWithIcon(UNSUCCESSFUL, error.message)
+			);
+
+			if (res && res.sigmaDaosProposalFilter) {
```
Megha-Dev-19:
```suggestion
			if (res?.sigmaDaosProposalFilter) {
```

src/pages/Proposal.vue
```diff
@@ -160,27 +191,49 @@ import { getProposalLsig, getDaoFundLSig } from "../contract/dao";
...
 			key: "ProposalKey",
+			DaoStore,
 		};
 	},
 	setup() {
 		const formState = reactive(ProposalStore());
 		const walletStore = reactive(WalletStore());
+		const proposalStore = reactive(ProposalTableStore());
 		const daoStore = reactive(DaoID());
+		const visible = ref<boolean>(false);
```
Megha-Dev-19:
```suggestion
		const isModalVisible = ref<boolean>(false);
```

src/styles/allDao.css
```diff
@@ -1,4 +1,4 @@
-.dao_table_container {
+.tableContainer {
```
Megha-Dev-19:
```suggestion
.table_container {
```

Megha-Dev-19 (changes requested):
Reviewed, I see that most of the proposal component looks same as AllDao, so can we add a common reusable component, as that would help in future if we have more such pages

src/components/AppHeader.vue
rmeena840: Bcz proposal list can be shown without actually opting. User will be redirected on proposal list on select or create dao, no matter address is selected or not.

src/components/AppHeader.vue
rmeena840: There was a bug in this. Whenever you change the input in the search field, the other tabs gets active. Now they will be active only when dao is selected. I had to add `searchDaoId` because input change needs to be reactive.

src/components/AppHeader.vue
rmeena840: No. This was also a bug. On selecting dao from the table, you will be able to see dao name but on searching, you won't be able to see it.

src/components/ProposalTable.vue
rmeena840: Close is more appropriate. I will change it to Close.

src/components/ProposalTable.vue
rmeena840: There is no type here.

src/pages/Proposal.vue
rmeena840: Done

src/styles/allDao.css
rmeena840: DOne

src/components/AppHeader.vue
Megha-Dev-19: but then we should add a check to select and close proposal

src/components/AppHeader.vue
Megha-Dev-19: why is that, if user search a correct dao id, all information regarding that should be visible, why to restrict it only when it is selected from dao table?

src/components/AppHeader.vue
Megha-Dev-19: okay, so now we have two daoID one that is searched and one which exists(according to user searched dao id), may be we can have two different state understand the difference

src/components/ProposalTable.vue
Megha-Dev-19: `"lll"` is the format type

src/components/AppHeader.vue
rmeena840: Since we will be storing new app (if found) in same state, then I would suggest to use one state.

src/components/AppHeader.vue
rmeena840: This is now happening on both. Earlier it was just on selecting dao but now also on search

src/components/ProposalTable.vue
rmeena840: Done

src/components/ProposalTable.vue
rmeena840: Done

src/components/AppHeader.vue
rmeena840: Close proposal will be handled in another task.

src/components/AppHeader.vue
Megha-Dev-19: okay got it, it was a bug before which you now fixed, good thanks

src/components/AppHeader.vue
Megha-Dev-19: okay then we will add the check in that task, works

Megha-Dev-19 (approved):
Thanks for the updates,LGTM

PR #63, opened 2022-07-26
Improvements

Megha-Dev-19:
- Updated names of few functions:
   `optInDaoApp` => `optInToAppUsingSecretKey`
   `optInToApp` => `optInToAppUsingLogicSig`
- Disabled Add Proposal if address is not selected
- Added `optInDaoApp` in Add Proposal.
- Added description in deposit token.
- Added minimum deposit amount from dao app global state.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-	"name": "sigmadao-ui",
+	"name": "SigmaDAO",
 	"description": "Sigma DAO UI",
 	"homepage": "https://algobuilder.dev",
 	"license": "mpl-2.0",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-	"name": "SigmaDAO",
+	"name": "sigma-dao",
 	"description": "Sigma DAO UI",
 	"homepage": "https://algobuilder.dev",
 	"license": "mpl-2.0",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-	"name": "sigma-dao",
+	"name": "SigmaDAO",
 	"description": "Sigma DAO UI",
 	"homepage": "https://algobuilder.dev",
 	"license": "mpl-2.0",
--- a/src/UIKit/TablePagination.vue
+++ b/src/UIKit/TablePagination.vue
@@ -34,10 +34,15 @@
 import { ROWS_PER_PAGE } from "@/constants";
 import { PaginationCallType } from "@/types";
 import { defineComponent, ref } from "vue";
+import { LeftOutlined, RightOutlined } from "@ant-design/icons-vue";
 
 export default defineComponent({
 	name: "TablePagination",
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "sigmadao-ui",
+	"name": "sigma-dao",
```
rmeena840:
```suggestion
	"name": "SigmaDAO",
```

src/pages/Proposal.vue
```diff
@@ -237,6 +244,29 @@ export default defineComponent({
 		toggleModalVisible() {
 			this.isModalVisible = !this.isModalVisible;
 		},
+		async optInDaoApp() {
+			try {
+				if (this.daoStore.dao_id) {
+					const isOptIn = await isApplicationOpted(
```
rmeena840:
```suggestion
					const isOptedIn = await isApplicationOpted(
```

src/pages/Proposal.vue
```diff
@@ -237,6 +244,29 @@ export default defineComponent({
...
+						this.walletStore.address,
+						this.daoStore.dao_id
+					);
+					if (!isOptIn) {
```
rmeena840:
```suggestion
					if (!isOptedIn) {
```

src/pages/Proposal.vue
```diff
@@ -325,7 +360,7 @@ export default defineComponent({
 								sk: new Uint8Array(0),
 							},
 							toAccountAddr: getApplicationAddress(this.daoStore.dao_id),
-							amount: 15,
+							amount: globalStateMinAmount as number,
```
rmeena840, line 400: Good catch!!

rmeena840 (changes requested):
Reviewed

src/UIKit/TablePagination.vue
```diff
@@ -34,10 +34,15 @@
 import { ROWS_PER_PAGE } from "@/constants";
 import { PaginationCallType } from "@/types";
 import { defineComponent, ref } from "vue";
+import { LeftOutlined, RightOutlined } from "@ant-design/icons-vue";
```
rmeena840, line 37: Imported but not using them???

package.json
Megha-Dev-19: There are certain provision regarding naming convention, using this throwed me the error, will see if there is any solution to it

src/UIKit/TablePagination.vue
Megha-Dev-19: Wasn't able to see the next, prev icons in pagination, when debugged I found the issue, so fixed it
LeftOutlined -> <left-outlined/>
RightOutlined -> <right-outlined/>

package.json
rmeena840: Pinged you a better approach

rmeena840 (approved):

PR #64, opened 2022-07-28
Added execute before input in add proposal 

Megha-Dev-19:


src/pages/Proposal.vue
```diff
@@ -237,6 +250,29 @@ export default defineComponent({
 		toggleModalVisible() {
 			this.isModalVisible = !this.isModalVisible;
 		},
+		//  execute_before must be after voting_end
```
rmeena840, line 253: Instead of doing this, can we disable all the day and time on or before voting_end and just allow the selection date ahead of voting_end?

rmeena840 (changes requested):
Reviewed

src/pages/Proposal.vue
Megha-Dev-19: I think that would be less intuitive for user, since he wouldn't understand why the dates are disabled, also what if the user add execute input before the voting input, it wouldn't work that way

src/pages/Proposal.vue
```diff
@@ -237,6 +250,29 @@ export default defineComponent({
...
+		async validateExecuteBefore(_rule: Rule, value: string) {
+			if (value === null) {
+				return Promise.reject("Please input execute before time.");
+			} else {
+				if (
+					this.formState.execute_before !== undefined &&
+					this.formState.vote_date !== undefined &&
+					this.formState.vote_date[1].length
```
rmeena840: I think you meant `this.formState,vote_date.length`. please check

rmeena840 (approved):
Left a comment

src/pages/Proposal.vue
Megha-Dev-19: no, its [1], to see if end date is added, [0] is start date, but I think I can update it with 0 as well

PR #65, opened 2022-08-01
Add: Use specific proposal and store proposal address in pinia

rmeena840:


src/store/DaoID.ts
```diff
@@ -6,6 +6,7 @@ export default defineStore("DaoStore", {
 		return {
 			dao_id: undefined,
 			govt_id: undefined,
+			proposal_addr: undefined,
```
Megha-Dev-19, line 9: I think we can remove the key initializer of the values which are undefined, since through type declarations. they already are undefined until provided with a value. What do you think.?

Megha-Dev-19 (approved):
LGTM, left a comment, Just want to confirm, this proposal address is the Lsig account address of proposal right?

src/store/DaoID.ts
rmeena840: No, we can't do that. If it contains a value then it won't be undefined until we explicitly set the value to undefined.

PR #66, opened 2022-08-01
Lsig Fund Check

Megha-Dev-19:
- Backend of account info added, fund check added

### DIFF
--- a/src/pages/Proposal.vue
+++ b/src/pages/Proposal.vue
@@ -196,8 +196,7 @@ import { types } from "@algo-builder/web";
 import type { LogicSigAccount } from "algosdk";
 import { getProposalLsig, getDaoFundLSig } from "../contract/dao";
 import { fundAmount, convertToSeconds, optInToApp } from "../utility";
-import { APP_NOT_FOUND, TOKEN_NOT_FOUND } from "@/constants";
-import { isApplicationOpted } from "@/indexer";
+import { getAccountInfoByAddress, isApplicationOpted } from "@/indexer";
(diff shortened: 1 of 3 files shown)


src/pages/Proposal.vue
```diff
@@ -379,30 +371,44 @@ export default defineComponent({
...
-				}
-				if (typeof this.daoStore.govt_id === "undefined") {
-					this.error = TOKEN_NOT_FOUND;
-					return;
+				const response = await getAccountInfoByAddress(lsig.address());
+				let amount = null;
+				if (response) {
+					amount = response?.amount;
 				}
+				if (amount) return;
```
rmeena840, line 461: There has to be min ammount that is needed by lsig.
@vuvoth Can you confirm this please?

rmeena840 (approved):
Please check comments

src/pages/Proposal.vue
Megha-Dev-19: yes @vuvoth, can you tell me about it?

vuvoth: Yes, I think so and this should be number algo we use for transfer when we create proprosal @Megha-Dev-19. Btw we should check the gov token when transfer. Did we do it?

Megha-Dev-19: Like I talked to you about it, I will create a new PR for dao fund lsig

Megha-Dev-19: And I have added the check for gov token transfer

PR #68, opened 2022-08-03
Create DAO updates

Megha-Dev-19:
- Added Dao lsig fund in create dao
- Added opt in to app using dao lsig account

### DIFF
--- a/src/utility/algod.utility.ts
+++ b/src/utility/algod.utility.ts
@@ -30,7 +30,7 @@ export const fundAmount = async (
 	console.log("Funded: ", response);
 };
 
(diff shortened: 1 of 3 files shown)


src/utility/algod.utility.ts
```diff
@@ -30,7 +30,7 @@ export const fundAmount = async (
 	console.log("Funded: ", response);
 };
 
-export const optInToAppUsingLogicSig = async (
+export const optInUsingLogicSig = async (
```
robert-zaremba:
```suggestion
export const optInUsingLsig = async (
```

robert-zaremba (approved):

PR #71, opened 2022-08-06
Reusable component added

Megha-Dev-19:
- Added reusable component for vote token description.


src/components/TokenDescription.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+	<a-row>
+		<a-col :span="12" :offset="6">
+			<p>
+				In order to vote for proposals you have to deposit your tokens. This
```
rmeena840: This is not a reusable component. The best way is to store this text in some constant file and pass this text as props to this component. In this way, If there is any other new description we want to display then we just have to give text as props.

+ Also, you can rename the component as SigmDAODescription or anything you like.
+ You should move this in UIKit. That is where we are adding reusable components.

Megha-Dev-19: done

rmeena840 (approved):
Lgtm

PR #72, opened 2022-08-10
Add: SigmaDAO compatible with app table in indexer

rmeena840:


Megha-Dev-19 (approved):
So now only sigma dao apps are stored in the indexer?

rmeena840:
> So now only sigma dao apps are stored in the indexer?

Correct. We now have separate indexer for SigmaDAO.

PR #77, opened 2022-08-31
Feature: Proposal Details along with withdraw and execute funcationality

rmeena840:
Screenshot 2022-08-31 at 9 30 35 PM


src/types/interface.types.ts
```diff
@@ -66,6 +66,11 @@ export interface DaoTableData {
 export interface ProposalTableData {
 	key?: number;
 	proposal_addr?: string;
+	amount?: number;
+	from?: string;
+	hash_algo?: string;
+	reciepient?: string;
```
Megha-Dev-19:
```suggestion
	recipient?: string;
```

Megha-Dev-19 (approved):
Lgtm. Nice job

Megha-Dev-19:
Suggestion: 
- We can remove the select and close action from all proposal page and move it inside proposals details page, since select proposal doesn't make sense if all proposal related things are done inside. 
What do you think.?

PR #79, opened 2022-09-05
Proposal Info UI Update

Megha-Dev-19:
image
image
image
image


robert-zaremba (commented):
Thanks for screen shots.
Please make sure that `clear vote record` is disabled (and probably in grey) if the voting is still active.

@rmeena840 can you review to make sure that the code is well organized?

src/pages/ProposalInfo.vue
```diff
@@ -61,55 +74,215 @@
...
 import AddressCopyable from "@/UIKit/Address.vue";
 import DaoID from "@/store/DaoID";
+import moment from "moment";
+import VoteStore from "@/store/VoteStore";
+import ProposalStore from "@/store/ProposalStore";
+import WalletStore from "@/store/WalletStore";
+import { types } from "@algo-builder/web";
+import { getAccountAppLocalState, getAssetInformation } from "@/indexer";
+import InfoToolTip from "@/components/InfoToolTip.vue";
+import { checkProposalResult } from "@/utility/proposal.utility";
```
rmeena840: `@/utility` should be sufficient to import. Please check

src/pages/WithdrawAndClose.vue
```diff
@@ -0,0 +1,189 @@
...
+				}
+			} else {
+				openErrorNotificationWithIcon(
+					UNSUCCESSFUL,
+					"Only creator of the proposal can withdraw from the proposal."
+				);
+			}
+		},
+		async handleCloseProposal() {
+			const proposalLsig: LogicSigAccount = await getProposalLsig(
```
rmeena840, line 158: Let's do one thing. We are using this getProposalLsig almost every place. Let's directly use it inside the method instead of here. From handleCloseProposal we will pass the required param.

rmeena840 (changes requested):
check comments

Megha-Dev-19:
> Thanks for screen shots. Please make sure that `clear vote record` is disabled (and probably in grey) if the voting is still active.
> 
@robert-zaremba Yes it is just like we have in execute and withdrawal, it is disabled.

src/pages/WithdrawAndClose.vue
Megha-Dev-19: I don't think that would be the right approach, because we need it here for checking current address (to know if it is proposer address), and that is something specific to UI, so it should be here instead of common functions.
What do you think..?

rmeena840: ok

rmeena840 (approved):
lgtm

PR #81, opened 2022-09-08
MyAlgo Wallet cross origin frame error fix

Megha-Dev-19:
image
Fixes the above error

### DIFF
--- a/src/store/WalletStore.ts
+++ b/src/store/WalletStore.ts
@@ -12,6 +12,7 @@ import {
 	openErrorNotificationWithIcon,
 	openSuccessNotificationWithIcon,
 } from "@/constants";
+import { shallowRef } from "vue";
 
 export default defineStore("WalletStore", {
 	state: (): WalletStoreState => {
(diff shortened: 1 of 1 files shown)


rmeena840 (approved):

src/store/WalletStore.ts
```diff
@@ -32,7 +33,15 @@ export default defineStore("WalletStore", {
 			this.walletKind = walletType;
 		},
 		setWebMode(webMode: WebModeTypes) {
-			this.webMode = webMode;
+			// my algo wallet has issue with ref in vue
+			// for reference : https://github.com/randlabs/myalgo-connect/issues/71
+			let myalgo = shallowRef() as any;
+			if (this.walletKind === WalletType.MY_ALGO) {
+				myalgo = webMode;
```
rmeena840, line 41: You have not set webmode for this?? How are you planning to use webmode for this wallet?

rmeena840 (approved):
Pre-approved. Please check comment before merge.

src/store/WalletStore.ts
Megha-Dev-19: in the next line I m setting the webmode to myalgo, actually the error is because of ref in vue, which is fixed by this. my algo is also a webmode

PR #87, opened 2022-09-29
Feature: Config env in backend

rmeena840:


backend/.env.betanet
```diff
@@ -0,0 +1,13 @@
+# BetaNet backend env file
+
+# TODO: Update below config for betanet
```
Megha-Dev-19, line 3: isn't it already updated?

Megha-Dev-19 (approved):
Left a comment, LGTM

backend/.env.betanet
rmeena840: No, we haven't deployed our app on any network.

PR #88, opened 2022-10-07
Feature: env support in SigmaDAO frontend for all 4 network (MainNet, TestNet etc.)

rmeena840:
Feature: env support in SigmaDAO frontend for all 4 network (MainNet, TestNet etc.)


.env.betanet
```diff
@@ -0,0 +1,12 @@
+# BetaNet env file
+
+# TODO: Below variabls are of privatenet. Update it with betanet env variables
```
Megha-Dev-19, line 3: why there is TODO here?

.env.testnet
```diff
@@ -0,0 +1,12 @@
+# TestNet env file
+
+# TODO: Below variabls are of privatenet. Update it with testnet env variables
+VUE_APP_NETWORK_NAME = "testnet"
+VUE_APP_BACKEND_URL = "http://localhost:4000/api"
+VUE_APP_ALGOSDK_TOKEN = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+VUE_APP_ALGOSDK_URL = "http://localhost"
```
Megha-Dev-19, line 7: these are used in all env files, is there a way to declare it once and use it in all other files, since if we ever change the endpoint we need to update all the files

Megha-Dev-19 (commented):
Left few suggestions

.env.betanet
rmeena840: We have not deployed the app yet. The env variable values for betanet is going to be different.

.env.testnet
rmeena840: It won't be unique when SigmaDAO is deployed on the network.

rmeena840:
@Megha-Dev-19 Ready for another review.

Megha-Dev-19 (approved):

PR #92, opened 2022-11-24
Update: UI enhancement and bug fix

rmeena840:
Update: UI enhancement and bug fix

### DIFF
--- a/dapp/src/pages/ProposalDetailsTab.vue
+++ b/dapp/src/pages/ProposalDetailsTab.vue
@@ -24,7 +24,7 @@
 			tab="Execute Proposal"
 			force-render
 		>
-			<ExecuteProposal :updateStatus="getStatusOfProposal" />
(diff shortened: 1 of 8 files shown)


dapp/src/pages/ProposalDetailsTab.vue
```diff
@@ -24,7 +24,7 @@
 			tab="Execute Proposal"
 			force-render
 		>
-			<ExecuteProposal />
+			<ExecuteProposal :updateStatus="getStatusOfProposal" />
```
Megha-Dev-19: this is the code which wasn't working, we can remove this from here and executeTx

Megha-Dev-19 (approved):

dapp/src/pages/ProposalDetailsTab.vue
rmeena840: ok

PR #94, opened 2022-12-01
Update: Algo and ASA Transfer tests

rmeena840:
Update: Algo and ASA Transfer tests

### DIFF
--- a/contracts/test/failing-path.js
+++ b/contracts/test/failing-path.js
@@ -477,6 +477,13 @@ describe("DAO - Failing Paths", function () {
 			}, RUNTIME_ERR1009);
 		});
 
+		it("Should reject if fee paid by daoFundLsig for type == 1 (ALGO TRANSFER)", function () {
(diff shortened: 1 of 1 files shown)


contracts/test/failing-path.js
```diff
@@ -476,6 +476,44 @@ describe("DAO - Failing Paths", function () {
 				ctx.executeTx(executeProposalTx);
 			}, RUNTIME_ERR1009);
 		});
+
+		it("Should reject if fee paid by daoFundLsig for type == 1 (ALGO TRANSFER)", function () {
```
Megha-Dev-19, line 480: I don't understand the tests, should it just be to execute the algo transfer and asa transfer .?

rmeena840: fee shouldn't be paid by dao. It should be paid by user. That's the test here.

Megha-Dev-19 (approved):

PR #96, opened 2022-12-08
Dao Info Page

Megha-Dev-19:
image


contracts/package.json
```diff
@@ -6,7 +6,7 @@
 	"dependencies": {
 		"@algo-builder/algob": "^6.0.0",
 		"@algo-builder/web": "^6.0.0",
```
robert-zaremba: let's update to 7.0

robert-zaremba (approved):

contracts/package.json
Megha-Dev-19: yeah sure will do it

